PR #1, opened 2016-07-21
implement monitoring API

schiessle:
Implement OCS API to provide some useful system informations

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -19,21 +19,34 @@
  *
  */
 
-/**
- * Create your routes in here. The name is the lowercase name of the controller
- * without the controller part, the stuff after the hash is the method.
- * e.g. page#index -> OCA\ServerInfo\Controller\PageController->index()
- *
- * The controller class has to be registered in the application.php file since
- * it's instantiated in there
- */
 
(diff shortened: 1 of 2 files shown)


lib/Controller/ApiController.php
```diff
@@ -20,13 +20,19 @@
  */
 
 
-namespace OCA\ServerInfo;
+namespace OCA\ServerInfo\Controller;
```
rullzer: `\Controller` was missing :wink:

lib/Controller/ApiController.php
```diff
@@ -20,13 +20,19 @@
...
 
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\OCSController;
 
 class ApiController extends OCSController {
 
+
+	public function __construct($appName,
```
rullzer: It seems the AppFramework needs this.
We should fix it properly. But for now this workaround does the job.

rullzer: PR to fix this https://github.com/nextcloud/server/pull/509

schiessle:
merged so that @Ivansss and I can work together on the same code basis

PR #5, opened 2016-08-01
add initial web UI

Ivansss:
This is an initial web UI for the system info app.
web-ui
web-ui_2


templates/part.content.php
```diff
@@ -1,10 +1,39 @@
...
+	<br>
+	<p><?php p($l->t('Memory'));?></p>
+	<canvas id="memorycanvas" width="400" height="100"></canvas>
+	<p><em id="memFooterInfo"></em></p>
+</div>
+<div class="section" id="storageSection">
+	<h2><?php p($l->t('Storage'));?></h2>
+	<p><?php p($l->t('Users: '));?><em id="numUsersStorage">--</em></p>
+	<p><?php p($l->t('Files: '));?><em id="numFilesStorage">--</em></p>
+	<p><?php p($l->t('Storages: '));?></p>
```
LukasReschke: Translators will move the empty space at the beginning depending on language. I'd move the space out here of the translatable string.

PR #7, opened 2016-08-02
Modified info.xml <dependencies> tag

totalcaos:
Modified dependency min-version to version 8.0
Added dependency max-version at version 10.0

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -9,6 +9,6 @@
     <namespace>ServerInfo</namespace>
     <category>other</category>
     <dependencies>
-        <owncloud min-version="9.1" />
+        <owncloud min-version="8.0" max-version="10" />
     </dependencies>
 </info>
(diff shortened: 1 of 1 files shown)


nickvergessen:
I think we should use 9.2 for both on master, @schiessle

schiessle:
We release it with stable10, so I would keep min version at 9.1 and increase max version to 9.2 for master. 

Min version 8.0 makes no sense, because there is no Nextcloud 8 and max 10 doesn't make sense because we can't tell today if it will be compatible with 10 (whatever 10 will be, at the moment we keep it align with ownCloud, so it would mean that it woks with whatever ownCloud10 and Nextcloud X will look like, something we just can tell as of today)

totalcaos:
updated as per @schiessle's comments

schiessle:
thanks @totalcaos :+1:

PR #42, opened 2016-08-29
fix active users chart

### LINKED ISSUE #21: Reverse order of active users
web-ui2

The x-axis of active users goes from present to past. It should be the other way around as with all time-based graphs.

@Ivansss

### LINKED ISSUE #37: Active users graph is drawn partially outside of its window
Steps to reproduce

  1)  create 2 users and connect
  2) go to Admin/server info
  3)  Active users graph is drawn partially outside of its window (see picture below)

capture d ecran 2016-08-21 a 12 10 43

### LINKED ISSUE #38: Don't use a spline for active users
It may look more fancy, but active users are integrals, so I think there should be straight lines between each points.

Ivansss:
Reversed order of active users and used straight lines between points.
Fix #21 
Fix #37 
Fix #38 

active-users

### DIFF
--- a/js/script.js
+++ b/js/script.js
@@ -180,7 +180,7 @@
 
 	function updateActiveUsersStatistics (activeUsers) {
 
-		var activeusers_data = [activeUsers.last5minutes, activeUsers.last1hour, activeUsers.last24hours];
+		var activeusers_data = [activeUsers.last24hours, activeUsers.last1hour, activeUsers.last5minutes],
 			stepSize = 0;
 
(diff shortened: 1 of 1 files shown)


schiessle:
:+1:

jancborchardt:
Nice, thanks a lot @Ivansss!

We should then still tune the values a bit as mentioned inhttps://github.com/nextcloud/serverinfo/issues/21#issuecomment-239186370

@schiessle is it possible to get more data points, like one each per hour, or several times a day at least, for the time frame of a week or two? That would then be very useful for the visualization. :)

PR #71, opened 2016-12-21
Don't throw an error when /proc/meminfo is not readable

### LINKED ISSUE #68: Handle non-accessible/non-existent /proc more gracefully
Version: 11.0.0

My hosting provider is all-inkl.com. In that hosting environment I do not have access to /proc and therefore /proc/meminfo.

When accessing the "Server Info" page as admin, "CPU Load" shows
Load average: undefined (Last minute)
"Memory usage" shows
Memory info not available

In the log file I get flodded with dozens of messages like this
```
file_get_contents(/proc/meminfo): failed to open stream: Permission denied at /www/htdocs/xxxxx/nextcloud/apps/serverinfo/lib/SystemStatistics.php#71
```
It would be nice if nextcloud would handle a non-accessible /proc more gracefully, and not floods the log file.
Also, the "Load average: undefined" message should be improved in such a case

nickvergessen:
Fix #68 

@schiessle

### DIFF
--- a/lib/SystemStatistics.php
+++ b/lib/SystemStatistics.php
@@ -68,7 +68,7 @@ class SystemStatistics {
 	 * @return array with the two values 'mem_free' and 'mem_total'
 	 */
 	protected function getMemoryUsage() {
-		$memoryUsage = file_get_contents('/proc/meminfo');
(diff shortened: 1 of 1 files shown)


schiessle:
@nickvergessen do you want to prepare a backport?

PR #76, opened 2017-01-04
[stable11] Don't throw an error when /proc/meminfo is not readable

nickvergessen:
Backport of #71


nickvergessen:
arg @schiessle this slipped through,
care to merge?

PR #83, opened 2017-02-08
Use is_readable() to check if meminfo can be read

patschi:
IMHO it's a bit more elegant way to check if meminfo can be read using is_readable(), instead of suppressing all warnings and errors for file_get_contents().


patschi:
Re-created as #93 with Signed-off-by.

PR #93, opened 2017-05-26
Use is_readable() to check if meminfo can be read

patschi:
IMHO it's a bit more elegant way to check if meminfo can be read using is_readable(), instead of suppressing all warnings and errors for file_get_contents(). (@ is changing the whole error_reporting() temporarly, so it should be avoided)

Re-created pull request from #83, now with 'Signed-off-by' line.


schiessle:
makes sense 👍

PR #96, opened 2017-06-01
Fixed phpMemLimit not being humanReadable-ized

patschi:
The display of the current PHP memory_limit on the info site did not get converted to a human readable value. That's fixed now.


schiessle:
Yes, this makes it look much better and easier to read 👍

PR #97, opened 2017-06-01
Checking for valid CPU average values

### LINKED ISSUE #90: Papercuts
During a review session we figured out some paper cuts.

  - [ ] CPU info - check for 0 - sys_getloadavg (this was on a UNIX afaik)
  - [ ] number of files might be wrong
  - [ ] change "files" to "files in data directory"
  - [ ] input field for external URL with copy link

This originated in https://github.com/nextcloud/server/issues/2672

cc @schiessle

patschi:
Added sanity-checks to validate return values by `sys_getloadavg()` before finally using them. This fixes errors on JavaScript side when it returns 0 (zero/null) or any other unexpected values.
Should solve #90 point 1.

### DIFF
--- a/js/script.js
+++ b/js/script.js
@@ -56,6 +56,9 @@
 			$('#cpuFooterInfo').text(t('serverinfo', 'CPU info not available'));
 			$('#cpuloadcanvas').addClass('hidden');
 			return;
+
+		} else if ($("#cpuloadcanvas").hasClass('hidden')) {
+			$("#cpuloadcanvas").removeClass('hidden');
 		}
 
 		var cpu1 = cpuload[0],
@@ -84,6 +87,9 @@
 			$('#memFooterInfo').text(t('serverinfo', 'Memory info not available'));
 			$('#memorycanvas').addClass('hidden');
 			return;
+
+		} else if ($("#memorycanvas").hasClass('hidden')) {
+			$("#memorycanvas").removeClass('hidden');
 		}
 
 		var memTotalBytes = memTotal * 1024,


patschi:
cc @schiessle @nickvergessen May anyone check this if it's okay?

lib/SystemStatistics.php
```diff
@@ -94,4 +95,25 @@ protected function getMemoryUsage() {
...
+	 */
+	protected function getProcessorUsage() {
+		// get current system load average.
+		$loadavg = sys_getloadavg();
+
+		// check if we got any values back.
+		if (!(is_array($loadavg) && count($loadavg) === 3)) {
+			// either no array or too few array keys.
+			// returning back zeroes to prevent any errors on JS side.
+			$loadavg = [0, 0, 0];
```
schiessle: wonder if it would make more sense to return a error code and then hide the CPU Load in the web interface or display something like "not available". Otherwise it looks like everything is OK and the system load is just super low, something which makes probably every admin happy. 😉

lib/SystemStatistics.php
```diff
@@ -94,4 +95,25 @@ protected function getMemoryUsage() {
...
+		}
+
+		return [
+			'loadavg' => $loadavg
```
schiessle: why do you put this in another array and don't return `$loadavg` directly?

lib/SystemStatistics.php
```diff
@@ -94,4 +95,25 @@ protected function getMemoryUsage() {
 		];
 	}
 
+	/**
+	 * Get current CPU load average
+	 *
+	 * @return array load average with three values, 1/5/15 minutes average.
+	 */
+	protected function getProcessorUsage() {
```
schiessle: Think this would be a nice method to write a small unit test.

lib/SystemStatistics.php
patschi: Done! Similar as when failing to retrieve memory usage.
image

lib/SystemStatistics.php
patschi: To allow easier extending the information without breaking something on the JS side. To be more extensible for the future. when extending stats with more CPU graph lines someday (e.g. splitting up CPU usage in user, system, kernel or something like that). But I can remove array if you want.

patschi:
As requested I've added warning on the client-side if CPU average couldn't retrieved on the server-side.

MorrisJobke:
I will move this to 15 so @schiessle can have a look at this.

Issue #86, opened 2017-03-19
Add information about outstanding app updates


janvonde:
To monitor the availability of app updates it would be nice to have this information in the api/v1/info endpoint.

patschi:
I think that's a good idea to monitor outstanding app updates from different monitoring solutions. I have done that right now, using that kind of format:
![image](https://user-images.githubusercontent.com/2029878/27011108-246b43cc-4eb5-11e7-9fa1-03e9c3c5e795.png)

It provides easy numeric overview, additionally you are able to parse all available apps which has updates available, including the new version number. I've tried multiple ways, but I think this one tells the most required information while keeping the output as few and clean as possible.

How's the data format from the monitoring perspective @janvonde? If it's okay for @schiessle or @nickvergessen a PR would be ready.

nickvergessen:
Looks good to me
I don't think there is an official standard for something like this, so go ahead

patschi:
Okay, fine! PR created.

PR #99, opened 2017-06-12
Added apps updates monitoring

patschi:
Adds apps updates monitoring to API endpoint, as requested in #86


lib/SystemStatistics.php
```diff
@@ -99,4 +110,41 @@ protected function getMemoryUsage() {
...
+		// sekeleton about the data we return back
+		$info = [
+			'num_installed' => 0,
+			'num_enabled' => 0,
+			'num_updates_available' => 0,
+			'app_updates' => [],
+		];
+
+		// load all apps
+		$apps = (new OC_App())->listAllApps();
```
MorrisJobke: Could you use `\OCP\IAppManager::getInstalledApps()` instead? (https://github.com/nextcloud/server/blob/591e75df5c3acf51e6968f20b1856481ee56f4de/lib/public/App/IAppManager.php#L116-L116) It's a list of app ids. And then also the CI job succeeds, because you don't use private namespace anymore ;)

patschi:
Thanks a lot for your feedback! Was searching quite a while for the right way to get out all installed apps... couldn't find any. As I found the exact way, before my recent commit, in an other app, I've just used the same way.

However, builds seems passing. Actually I got into one new issue that I was not able to find out the correct way to get all activated apps at all. Using `IAppManager` I just found the following and closest ways:
 - `getAlwaysEnabledApps()`, which are forced-online apps which are required like `files` app
 - and `getEnabledAppsForUser(IUser $user)` for  getting activated apps depending on the user. But the monitoring endpoint is userless?

So I have just removed `num_enabled` count for now. Either we'll find a better way and implement that, or we just add it someday later back.

MorrisJobke:
> So I have just removed num_enabled count for now. Either we'll find a better way and implement that, or we just add it someday later back.

Okay for me for now 👍

tflidd:
What is the status of this PR? There are already conflicts preventing to merge this. Would be great to know if this is a wanted change that we can resolve the conflicts and merge it in. It's open for nearly a year.

MorrisJobke:
Makes sense to add it IMO 👍 

cc @blizzz @rullzer

rullzer:
There is a conflict.
But in general sure lets do this!

patschi:
Conflict resolved.
@MorrisJobke @rullzer 

Probably also worth checking other PRs, like #98 and #97?

lib/SystemStatistics.php
```diff
@@ -24,25 +24,38 @@
...
+use OC\App\AppStore\Fetcher\AppFetcher;
 
 class SystemStatistics {
 
 	/** @var IConfig */
 	private $config;
-
 	/** @var View view on data/ */
 	private $view;
+	/** @var appFetcher */
```
nickvergessen: `AppFetcher`

lib/SystemStatistics.php
```diff
@@ -24,25 +24,38 @@
...
+	private $appFetcher;
+	/** @var appManager */
```
nickvergessen: `IAppManager`

lib/SystemStatistics.php
```diff
@@ -24,25 +24,38 @@
...
+	private $appManager;
 
 	/**
 	 * SystemStatistics constructor.
 	 *
-	 * @param IConfig $config
+ 	 * @param IConfig $config
+	 * @param IAppManager $appManager
 	 */
```
nickvergessen: remove the doc block, just duplicating the method signature anyway.
Or you add the `$appFetcher`

tflidd:
You probably want to merge this branch into master (not the other way round like your did in 13e5bd2

PR #109, opened 2017-12-24
Show swap usage in FreeBSD systems

### LINKED ISSUE #108: Error in FreeBSD (Swap)
On NC13beta1.

Use serverinfo on a FreeBSD system. 

`{"reqId":"QKrCFayFmbKiLz4RuNxG","level":3,"time":"2017-12-10T22:06:44+00:00","remoteAddr":"2001:::","user":"user","app":"PHP","method":"GET","url":"\/index.php\/apps\/serverinfo\/update","message":"Undefined index: swap_total at \/usr\/local\/www\/apache24\/data\/nextcloud\/apps\/serverinfo\/lib\/SystemStatistics.php#62","userAgent":"Mozilla\/5.0 (X11; Ubuntu; Linux x86_64; rv:57.0) Gecko\/20100101 Firefox\/57.0","version":"13.0.0.6"}`

https://github.com/nextcloud/serverinfo/pull/105 introduced a separate curve for swap usage which is not read yet from FreeBSD systems.

quick workaround: just send `0` values or `N/A` for swap
real solution: implement swap read-out on FreeBSD

tflidd:
Fixes #108 

Tested on FreeBSD 11 + apache24 + php 7.1

small C program to eat up your RAM slowly (for testing): https://www.linuxatemyram.com/play.html

Signed-off-by: Joachim Boerner <tflidd@aspekte.net>

### DIFF
--- a/lib/SystemStatistics.php
+++ b/lib/SystemStatistics.php
@@ -77,13 +77,27 @@ class SystemStatistics {
 		}
 		//If FreeBSD is used and exec()-usage is allowed
 		if (PHP_OS === 'FreeBSD' && \OC_Helper::is_function_enabled('exec')) {
+			//Read Swap usage:
+			exec("/usr/sbin/swapinfo",$return,$status);
+			if ($status===0 && count($return) > 1) {
+				$line = preg_split("/[\s]+/", $return[1]);			
+				if(count($line) > 3) {
+					$swapTotal = (int) $line[3];
+					$swapFree = $swapTotal- (int) $line[2];
+				}
+			}
+			unset($status);
+			unset($return);
+			//Read Memory Usage
 			exec("/sbin/sysctl -n hw.physmem hw.pagesize vm.stats.vm.v_inactive_count vm.stats.vm.v_cache_count vm.stats.vm.v_free_count",$return,$status);
 			if ($status===0) {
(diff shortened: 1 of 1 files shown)


lib/SystemStatistics.php
```diff
@@ -77,13 +77,27 @@ protected function getMemoryUsage() {
...
+			//Read Swap usage:
+			exec("/usr/sbin/swapinfo",$return,$status);
+			if ($status===0 && count($return) > 1) {
+				$line = preg_split("/[\s]+/", $return[1]);			
+				if(count($line) > 3) {
+					$swapTotal = (int) $line[3];
+					$swapFree = $swapTotal- (int) $line[2];
+				}
+			}
+			unset($status);
```
MorrisJobke: Why is there two `unset()`? Those should not be needed and also they would cause the if case in line 93 to never be true.

tflidd: I want to use them again in line 93 (it's the output of the `exec()` command). And I wanted to avoid that in case it fails it keeps the previous values and does strange things. I could have used new variables in line 93 if that is preferred.

MorrisJobke: A missed that - then it's fine 👍

PR #126, opened 2018-09-01
Make appinfo/info.xml NC14-compliant

patschi:
This PR should fix following error of certain CI tasks by providing a `<summary>` tag within the `info.xml` file:
```
Invalid appinfo.xml file found: Element 'description': This element is not expected. Expected is one of ( name, summary ).
```

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0"?>
-<info>
+<info xmlns:xsi= "http://www.w3.org/2001/XMLSchema-instance"
+	  xsi:noNamespaceSchemaLocation="https://apps.nextcloud.com/schema/apps/info.xsd">
     <id>serverinfo</id>
     <name>Monitoring</name>
     <summary>Monitoring app with useful server information</summary>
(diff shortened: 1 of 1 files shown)


appinfo/info.xml
```diff
@@ -2,19 +2,21 @@
 <info>
```
nickvergessen: Replace this with:
```xml
<info xmlns:xsi= "http://www.w3.org/2001/XMLSchema-instance"
      xsi:noNamespaceSchemaLocation="https://apps.nextcloud.com/schema/apps/info.xsd">
```

It helps IDEs and other tools to validate the info.xml automatically

nickvergessen:
Looks good apart from the little issue above

nickvergessen:
btw @patschi if you push your branches to out git repo directly, we can collaborate on the branches directly to fix minor issues like this.

You should have permissions to push to our repos already :nerd_face:

patschi:
Good point, @nickvergessen! Will do! :)

nickvergessen:
Seems like stable14 was branched off before your PRs.
Can you create PRs for all your issues to fix them in stable14 as well?

PR #127, opened 2018-09-02
Add "Nextcloud" section to serverinfo UI page

patschi:
This adds a few monitoring information we already collect anyway - and were already available in the API endpoint.

This currently loads all SystemStatistics, which might be reduced to a smaller dataset in the future for performance reasons.

How it looks like:
image

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -1,20 +1,23 @@
 <?xml version="1.0"?>
-<info>
+<info xmlns:xsi= "http://www.w3.org/2001/XMLSchema-instance"
+	  xsi:noNamespaceSchemaLocation="https://apps.nextcloud.com/schema/apps/info.xsd">
     <id>serverinfo</id>
     <name>Monitoring</name>
+    <summary>Monitoring app with useful server information</summary>
     <description>Provides useful server information, such as CPU load, RAM usage, disk usage, number of users, etc.</description>
-    <licence>AGPL</licence>
+    <version>1.4.0</version>
+    <licence>agpl</licence>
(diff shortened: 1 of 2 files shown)


templates/settings-admin.php
```diff
@@ -53,6 +53,14 @@
 	<h2><?php p($l->t('Storage'));?></h2>
 	<p><?php p($l->t('Users:'));?> <em id="numUsersStorage"><?php p($_['storage']['num_users']);?></em></p>
 	<p><?php p($l->t('Files:'));?> <em id="numFilesStorage"><?php p($_['storage']['num_files']);?></em></p>
+	<p><?php p($l->t('Storages:'));?> <em id="numFilesStorages"><?php p($_['storage']['num_storages']);?></em></p>
+	<p><?php p($l->t('Free Space:'));?> <em id="numFilesFreeSpace"><?php p($_['system']['freespace']);?></em></p>
```
nickvergessen: `numFilesFreeSpace` is a weird name though,
I guess it should be `diskFreeSpace` because it's not related to the number of files

patschi: I've decided to go with `systemDiskFreeSpace` as it also comes from the `SystemStatistics` class

patschi:
@nickvergessen Unfortunately I'm not able to merge the PR myself due to this reason:
```
Required statuses must pass before merging
All required statuses and check runs on this pull request must run successfully to enable automatic merging.
```

And at least one check will fail anyway, due to the `OC_Helper` thing. I assume you're having more permission to force the merge on this.

PR #128, opened 2018-09-02
Fix static call of OC_Helper

patschi:
Fixes static call of OC_Helper, causing the app not being compliant:
```
# sudo -u www-data php /var/www/nextcloud/occ app:check-code serverinfo
Analysing /var/www/nextcloud/apps/serverinfo/lib/SystemStatistics.php
 1 errors
    line   96: OC_Helper - Static method of private class must not be called
```

Together with PR #126 the app is now fully compliant, based on testing on NC14 RC2:
```
# sudo -u www-data php /var/www/nextcloud/occ app:check-code serverinfo
App is compliant - awesome job!
```

### DIFF
--- a/lib/SystemStatistics.php
+++ b/lib/SystemStatistics.php
@@ -195,14 +195,10 @@ class SystemStatistics {
 		if (!function_exists($function_name)) {
 			return false;
 		}
-		$disabled = explode(',', $this->phpIni->get('disable_functions') ?: '');
-		$disabled = array_map('trim', $disabled);
-		if (in_array($function_name, $disabled)) {
+		if ($this->phpIni->listContains('disable_functions', $function_name)) {
 			return false;
 		}
-		$disabled = explode(',', $this->phpIni->get('suhosin.executor.func.blacklist') ?: '');
-		$disabled = array_map('trim', $disabled);
-		if (in_array($function_name, $disabled)) {
+		if ($this->phpIni->listContains('suhosin.executor.func.blacklist', $function_name)) {
 			return false;
 		}
 		return true;
--- a/lib/SystemStatistics.php
+++ b/lib/SystemStatistics.php
@@ -26,6 +26,7 @@ use OC\Files\View;
 use OC\Installer;
 use OCP\IConfig;
 use OCP\App\IAppManager;
+use bantu\IniGetWrapper\IniGetWrapper;
 
 class SystemStatistics {
 
@@ -37,6 +38,8 @@ class SystemStatistics {
 	private $appManager;
 	/** @var Installer */
 	private $installer;
+    /** @var IniGetWrapper */
+    protected $phpIni;
 
 	/**
 	 * SystemStatistics constructor.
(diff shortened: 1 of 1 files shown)


danielkesselberg:
Is this a valid way? You are calling a static method like a non-static method (https://github.com/nextcloud/server/blob/2e36069e24406455ad3f3998aa25e2a949d1402a/lib/private/legacy/helper.php#L475)

nickvergessen:
> bildschirmfoto von 2018-09-03 12-43-44

Well, it is "allowed" in php, but a sane IDE complains about it.
It also just "breaks" the checker instead of solving the real problem. Either `is_function_enabled` should be moved to the OCP\Utils, or it should not be used by apps.

This is exactly why the check was added :wink: 

@rullzer any opinion? Make the method public, disable the checker in this app or copy the method content to this app.

patschi:
IMHO I'd say using separated `is_function_enabled()` function within the app for now is a good idea. This way the concrete details might be worked out in the future.
* Adding this function to `OCP\Utils` just a few days before the NC 14 stable release might not be the best idea. This way we can provide a compliant `serverinfo` app while this is getting solved/changed.
* Theoretically the app should also run fine on NC13 as well.

lib/SystemStatistics.php
```diff
@@ -179,4 +184,28 @@ protected function getAppsInfo() {
...
+	 *
+	 * @param string $function_name
+	 * @return bool
+	 */
+	public function is_function_enabled($function_name) {
+		if (!function_exists($function_name)) {
+			return false;
+		}
+		$disabled = explode(',', $this->phpIni->get('disable_functions') ?: '');
+		$disabled = array_map('trim', $disabled);
```
danielkesselberg: I would like to suggest a small change here. IniGetWrapper provide a [listContains](https://github.com/bantuXorg/php-ini-get-wrapper/blob/master/src/IniGetWrapper.php#L120) function and does exactly what 198-202 does.

`if ($this->phpIni->listContains('disable_functions', $function_name))` should do the trick.

lib/SystemStatistics.php
```diff
@@ -179,4 +184,24 @@ protected function getAppsInfo() {
...
+	 * @return bool
+	 */
+	public function is_function_enabled($function_name) {
+		if (!function_exists($function_name)) {
+			return false;
+		}
+		if ($this->phpIni->listContains('disable_functions', $function_name)) {
+			return false;
+		}
+		if ($this->phpIni->listContains('suhosin.executor.func.blacklist', $function_name)) {
```
danielkesselberg: Thank you! May i come up with another suggestion? I was wondering what happened to suhosin. Latest release is only available for php 5.6. There is another version for php 7.x (https://github.com/sektioneins/suhosin7) but they never released a production ready version. I guess there is only a little change that someone is using php7 with suhosin.

@nickvergessen do we need the check for suhosin.executor.func.blacklist?

patschi: I was wondering about the exact same just right before when following your suggestion. According to most recent [System Requirements](https://docs.nextcloud.com/server/14/admin_manual/installation/system_requirements.html) of NC14 the runtime PHP 5.6 is still listed as supported, so I decided to keep the few lines for now.

However apps like `oauth2` or `theming` already takes advantage of the `Strict type declaration` which is available starting PHP 7 and up...

danielkesselberg: fyi https://github.com/nextcloud/server/issues/7368

nickvergessen: NC14 requires php7+, the docs need adjustments

Im fine with leaving out the suhosin here

patschi: OK. Removed then.

PR #125, opened 2018-09-01
Fix static call for isUpdateAvailable

danielkesselberg:
Close https://github.com/nextcloud/server/issues/10950

PR #132, opened 2018-09-05
Backport  master to stable14

danielkesselberg:
Ref #126 
Ref #125 
Ref #128


nickvergessen:
Normally we don't do it like this but instead backport the individually PRs.
but i guess its also fine atm

PR #146, opened 2019-03-05
adding more info to the page. This is only the first step

karlitschek:
Signed-off-by: Frank Karlitschek <frank@nextcloud.com>


lib/OperatingSystems/DefaultOs.php
```diff
@@ -0,0 +1,168 @@
...
+		$uptime = shell_exec('uptime -p');
+		return $uptime;
+	}
+
+
+	/**
+	* @return string
+	*/
+	public function getTimeServers() {
+		$uptime = shell_exec('cat /etc/ntp.conf |grep  \'^pool\' | cut -f 2 -d " "');
```
patschi: I see this just a bit problematic for systems using systemd, like major server distributions Debian and Ubuntu. In this cases the NTP servers would be in `/etc/systemd/timesyncd.conf` instead. So the question is how to handle that cases - simply using systemd-based checks here, or creating own `OperatingSystem` classes for systemd-systems and non-systemd-systems?

lib/OperatingSystems/DefaultOs.php
```diff
@@ -0,0 +1,168 @@
...
+		return $uptime;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getNetworkInfo() {
+			$result=array();
+			$result['hostname'] = \gethostname();
+			$dns = shell_exec('cat /etc/resolv.conf |grep -i \'^nameserver\'|head -n1|cut -d \' \' -f2');
```
patschi: If I'm right this has also changed in Ubuntu 18.04 with systemd, where the integrated systemd-dns with entry `nameserver 127.0.0.53` is being used. But to be honest, I've no clue how to efficiently get the current DNS servers which are in use. According to [stackoverflow here](https://stackoverflow.com/questions/50299241/ubuntu-18-04-server-how-to-check-dns-ip-server-setting-being-used/50329033#50329033) we can get the current DNS server by parsing the output of `systemd-resolve --status`...

lib/OperatingSystems/DefaultOs.php
karlitschek: Yes. Unfortunately there are a lot more differences between different operating systems. There is a system in place to have different backends for the different distributions. At the moment this has only the default backend which should often works. Tested with Ubuntu here. But this requires more backends for different distros. As you described.

lib/OperatingSystems/DefaultOs.php
karlitschek: yes. we need a different backend for this case then.

lib/OperatingSystems/DefaultOs.php
MorrisJobke: > Tested with Ubuntu here.

Which one? Because systemd is already in place since quite some time and seems to be the de-facto standard, that has the most coverage easily.

patschi: Exactly. As I remember systemd is in place since Ubuntu 16.04 (first release on April 21, 2016) and Debian 8 (first release on April 26, 2015). `FreeBSD` and sort of needs a special handling anyway - but that's another topic for later.

karlitschek: Great feedback. I'm reading this file now additionally. On my Ubuntu 18.10 the config is in /conf/ntp.conf Probably because this is an older install and the config is migrated from an old Ubuntu

karlitschek:
@skjnldsv ping about the theme color

karlitschek:
> @skjnldsv ping about the theme color

solved already :-)

js/script.js
```diff
@@ -354,10 +354,10 @@ $(document).ready(function(){
 					datasets : [
 						{
 							backgroundColor: [
-				                OCA.Theming.color,
-				                'rgb(249, 249, 249)',
-				            ],
-				            borderWidth: 0,
+								OCA.Theming.color,
```
skjnldsv: I just realised that theming can be disabled, so you'll have to check if OCA.Theming exists first and add a fallback color ;)

karlitschek: really? you mean disabling the theming app? How do you suggest to check for that in the correct way?

skjnldsv: > you mean disabling the theming app?

yes :)
I would just do a ternary check:
```js 
OCA.Theming
    ? OCA.Theming.color
    : 'rgb(54, 129, 195)'
```

karlitschek: ahh cool. this makes sense

lib/Controller/ApiController.php
```diff
@@ -34,6 +35,9 @@
 
 class ApiController extends OCSController {
 
+	/** @var Os */
+	private $Os;
```
MorrisJobke:
```suggestion
	private $os;
```

lib/Controller/ApiController.php
```diff
@@ -66,6 +71,7 @@ class ApiController extends OCSController {
 	 */
 	public function __construct($appName,
 								IRequest $request,
+								Os $Os,
```
MorrisJobke:
```suggestion
								Os $os,
```

lib/Controller/ApiController.php
```diff
@@ -75,6 +81,7 @@ public function __construct($appName,
 	) {
 		parent::__construct($appName, $request);
 
+		$this->Os = $Os;
```
MorrisJobke:
```suggestion
		$this->os = $os;
```

lib/Controller/ApiController.php
```diff
@@ -110,6 +117,25 @@ public function info() {
...
 	}
 
+	/**
+	 * @return DataResponse
+	 */
+	public function BasicData(): DataResponse {
+		$servertime = $this->Os->getTime();
+		$uptime = $this->Os->getUptime();
+		$timeservers = $this->Os->getTimeServers()[0];
+		return new DataResponse(array('servertime'=>$servertime,'uptime'=>$uptime,'timeservers'=>$timeservers));
```
MorrisJobke: use `[]` instead of `array()` and spaces around `=>`. Also spaces after `,`. Best would be to put every entry on a new line.

lib/OperatingSystems/DefaultOs.php
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getMemory() {
+		$memory = shell_exec('cat /proc/meminfo  | grep -i \'MemTotal\' | cut -f 2 -d ":" | awk \'{$1=$1}1\'');
+		$memory = explode(' ',$memory);
+		$memory = round($memory[0]/1024);
+		if ($memory<1024) {
```
MorrisJobke: spaces around operator

lib/OperatingSystems/DefaultOs.php
```diff
@@ -0,0 +1,169 @@
...
+			$memory = $memory.' MB';
```
MorrisJobke: spaces around `.` (same for the following lines)

lib/OperatingSystems/DefaultOs.php
```diff
@@ -0,0 +1,169 @@
...
+		return $memory;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getCPUName() {
+		$cpu   = shell_exec('cat /proc/cpuinfo  | grep -i \'Model name\' | cut -f 2 -d ":" | awk \'{$1=$1}1\'');
+		$cores = shell_exec('cat /proc/cpuinfo  | grep -i \'cpu cores\' | cut -f 2 -d ":" | awk \'{$1=$1}1\'');
+		if ($cores == 1) $cores = ' ('.$cores.' core)'; else $cores = ' ('.$cores.' cores)';
```
MorrisJobke: Always put curly brackets around if blocks and put the blocks on new lines.

lib/OperatingSystems/DefaultOs.php
```diff
@@ -0,0 +1,169 @@
...
+		foreach($interfaces as $interface) {
+				$iface = array();
+				$iface['interface'] = basename($interface);
+				$iface['mac']  = shell_exec('ip addr show dev '.$iface['interface'].' | grep "link/ether " | cut -d \' \' -f 6  | cut -f 1 -d \'/\'');
+				$iface['ipv4'] = shell_exec('ip addr show dev '.$iface['interface'].' | grep "inet " | cut -d \' \' -f 6  | cut -f 1 -d \'/\'');
+				$iface['ipv6'] = shell_exec('ip -o -6 addr show '.$iface['interface'].' | sed -e \'s/^.*inet6 \([^ ]\+\).*/\1/\'');
+				if ($iface['interface']<>'lo') {
+					$iface['status'] = shell_exec('cat /sys/class/net/'.$iface['interface'].'/operstate');
+					$iface['speed'] = shell_exec('cat /sys/class/net/'.$iface['interface'].'/speed');
+					if ($iface['speed'] <> '') $iface['speed'] = $iface['speed'] . 'Mbps'; else $iface['speed'] = 'unknown';
```
MorrisJobke: Please use `!===` instead of `<>`, because that's what we use typically.

lib/OperatingSystems/DefaultOs.php
```diff
@@ -0,0 +1,169 @@
...
+			$result['gateway'] = $gw;
+			return $result;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getNetworkInterfaces() {
+		$interfaces = glob('/sys/class/net/*');
+		$result=array();
```
MorrisJobke:
```suggestion
		$result = [];
```

lib/OperatingSystems/DefaultOs.php
```diff
@@ -0,0 +1,169 @@
...
+	 * @return array
+	 */
+	public function getDiskInfo() {
+		$blacklist = array('','Type','tmpfs','devtmpfs');
+
+		$data = shell_exec('df -T');
+		$lines = preg_split('/[\r\n]+/', $data);
+		foreach ($lines as $line) {
+			$entry = preg_split('/\s+/', trim($line));
+			if(isset($entry[1]) and  !in_array($entry[1],$blacklist)) {
```
MorrisJobke: Please use `&&` instead of `and`.

lib/OperatingSystems/DefaultOs.php
```diff
@@ -0,0 +1,169 @@
...
+				$items = array();
```
MorrisJobke: use `[]`

lib/Os.php
```diff
@@ -0,0 +1,247 @@
...
+	 * @param IL10N $l
+	 */
+	public function __construct(IClientService $clientService, IConfig $config, IDBConnection $connection, IniGetWrapper $phpIni, IL10N $l) {
+		$this->clientService = $clientService;
+		$this->config = $config;
+		$this->connection = $connection;
+		$this->phpIni = $phpIni;
+		$this->l = $l;
+
+		$detectedOs = $this -> detectOs();
```
MorrisJobke:
```suggestion
		$detectedOs = $this->detectOs();
```

lib/Os.php
```diff
@@ -0,0 +1,247 @@
...
+		switch ($detectedOs) {
+			case 'ubuntu':
+				$this->backend = new DefaultOs();
+				break;
+			default:
+				$this->backend = new DefaultOs();
+				break;
+		}
```
MorrisJobke: This whole `switch` block doesn't make sense 🤔

lib/Os.php
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+
+	/**
+	 * @return string
+	 */
+	public function detectOs() {
+		$release = shell_exec('cat /etc/*-release');
+		if(stripos($release,'ubuntu')) {
```
MorrisJobke: Please use spaces after the `,`

lib/Os.php
```diff
@@ -0,0 +1,247 @@
...
+			return('unknown');
+		}
+	}
+
+
+	/**
+	 * @return bool
+	 */
+	public function supported() {
+		$data = $this -> backend -> supported();
```
MorrisJobke: Please don't use spaces around arrows.

lib/Os.php
```diff
@@ -0,0 +1,247 @@
...
+
+
+		$item['device'] = 'device3';
+		$item['fs'] = 'ext4';
+		$item['used'] = 10000000;
+		$item['available'] = 50000000;
+		$item['percent'] = '90%';
+		$item['mount'] = '/data3';
+		$data[] = $item;
+*/
```
MorrisJobke: This looks like debug output that should not be checked in.

lib/Os.php
```diff
@@ -0,0 +1,247 @@
...
+		$disks = $this -> backend -> getDiskInfo();
+		$data = array();
+		$i = 0;
+		foreach($disks as $disk){
+			$data[$i] = array(round(($disk['used'])/1024/1024,1),round($disk['available']/1024/1024,1));
+			$i++;
+		}
+
+//		debug data
+//		$data = array('0'=>array(1,2),'1'=>array(4,5),'2'=>array(3,1));
```
MorrisJobke: Debug data again.

lib/Os.php
```diff
@@ -0,0 +1,247 @@
...
+
+	/**
+	 * @return string
+	 */
+	public function getDiskData() {
+		$disks = $this -> backend -> getDiskInfo();
+		$data = array();
+		$i = 0;
+		foreach($disks as $disk){
+			$data[$i] = array(round(($disk['used'])/1024/1024,1),round($disk['available']/1024/1024,1));
```
MorrisJobke: Add spaces around operators and behind `,`

lib/Os.php
```diff
@@ -0,0 +1,247 @@
...
+				$this->backend = new DefaultOs();
+				break;
+		}
+	}
+
+
+	/**
+	 * @return string
+	 */
+	public function detectOs() {
```
MorrisJobke: use this:

```suggestion
	public function detectOs(): string {
```

instead of the doc block, if the doc block doesn't have more details.

templates/settings-admin.php
```diff
@@ -26,54 +26,189 @@
...
+			echo('</p><p>');
+			p($l->t('Size').': ');
+			echo('<span class="info">'.FormatBytes($disk['used']+$disk['available']).'</span>');
+			echo('</p><p>');
+			p($l->t('Available').': ');
+			echo('<span class="info">'.FormatBytes($disk['available']).'</span>');
+			echo('</p><p>');
+			p($l->t('Used').': ');
+			echo('<span class="info">'.$disk['percent'].'</span>');
+			echo('</p></div>');
```
MorrisJobke: It's maybe better to read to leave PHP and write plain HTML instead of the `echo`'s on every line.

templates/settings-admin.php
```diff
@@ -26,54 +26,189 @@
...
+					p($l->t('MAC').': ');
+					echo('<span class="info">'.$interface['mac'].'</span>');
+					echo('</p><p>');
+				}
+				p($l->t('IPv4').': ');
+				echo('<span class="info">'.$interface['ipv4'].'</span>');
+				echo('</p><p>');
+				p($l->t('IPv6').': ');
+				echo('<span class="info">'.$interface['ipv6'].'</span>');
+				echo('</p></div>');
```
MorrisJobke: Same here.

tests/lib/Controller/ApiControllerTest.php
```diff
@@ -1,137 +0,0 @@
...
-		$this->assertTrue(isset($data['nextcloud']));
-		$this->assertTrue(isset($data['server']));
-
-		$this->assertSame('systemStatistics', $data['nextcloud']['system']);
-		$this->assertSame('storageStatistics', $data['nextcloud']['storage']);
-		$this->assertSame('shareStatistics', $data['nextcloud']['shares']);
-		$this->assertSame('unknown', $data['server']['webserver']);
-		$this->assertSame('databaseStatistics', $data['server']['database']);
-		$this->assertSame('phpStatistics', $data['server']['php']);
-		$this->assertSame('sessionStatistics', $data['activeUsers']);
```
MorrisJobke: Maybe extend the tests instead of removing them. Then you can see when the API changes.

PR #175, opened 2020-02-26
Avoid line breaks after long device names in "df" command

### LINKED ISSUE #173: Command df parse error when the string `filesystem` is long
### Steps to reproduce
1. have a very long `filesystem` field in the output of `df`. e.g. by installing nextcloud via docker image
2. enter the docker image and execute df
图片
3. Enter the system info tab in the admin setting from the nextcloud web interface. 
图片

### Expected behaviour
Parse the df output correctly and the figure will show filesystem as `/dev/mapper/ubuntu--vg-root` and all other fields are corrected accordingly.

mikoMK:
Long device names cause unwanted line breaks after them in the `df` output. This leads to wrong parsing of the data:
2020-02-26 22_59_46-mRemoteNG - confCons xml - Nextcloud

The `-P` option fixes that (`-P, --portability / use the POSIX output format`):
2020-02-26 23_00_06-mRemoteNG - confCons xml - Nextcloud

Fixes #173

### DIFF
--- a/lib/OperatingSystems/DefaultOs.php
+++ b/lib/OperatingSystems/DefaultOs.php
@@ -154,7 +154,7 @@ class DefaultOs {
 	 */
 	public function getDiskInfo() {
 		$blacklist = ['', 'Type', 'tmpfs', 'devtmpfs'];
-		$data  = shell_exec('df -T');
+		$data  = shell_exec('df -TP');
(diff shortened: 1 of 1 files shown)


juliushaertl:
Thanks a lot for this :+1:

PR #179, opened 2020-03-05
Add the number of active users in the last 7 and 30 days

arnauvp:
We think it can be interesting to have active session data available for periods longer than 24 hours. We are adding here 7 and 30 days to the API. Do you think it would be interesting to include these values in the official Monitoring app? If so we could also tweak the charts to add them as part of the PR.


arnauvp:
@nickvergessen would you find this feature useful? Do you see any drawback?

tflidd:
I don't see any changes to the user chart. You have to add something here as well?
https://github.com/nextcloud/serverinfo/blob/e8e195f77f26939b6fdc62063a0c98ef3cd27fcb/js/script.js#L293

arnauvp:
Yes; we just use the API but for consistency it would be nice to show it in the charts as well. I'll see when I can allocate some time for it.

arnauvp:
OK we're pretty busy with other things so let me close this one and we'll reopen it if we ever find someone to continue it.

Issue #160, opened 2019-10-21
NTP servers not displayed


j-ed:
## Expected behavior
The NTP servers should be displayed.

## Current behavior
The NTP servers are not displayed.

![grafik](https://user-images.githubusercontent.com/2572764/67198678-c511d100-f3ff-11e9-8f27-f165b67a1ebc.png)

## Steps to reproduce
1. Open the systems app in the settings dialog.

I tried to find the root cause of the problem and identified an incorrect command line syntax causing the problem in `lib/OperatingSystems/DefaultOs.php`:
```
    public function getTimeServers() {
        $servers = shell_exec('cat /etc/ntp.conf |grep  \'^pool\' | cut -f 2 -d " "');
        $servers .= ' ' . shell_exec('cat /etc/systemd/timesyncd.conf |grep  \'^NTP=\' | cut -f 2 -d " "');
        return $servers;
    }
```
This assumes that the pool command is always being used in each `ntp.conf` file what isn't requured. In my configuration file only one server is given, e.g.
```
server <name-of-my-time-server>
```
BTW, I would recommend to change the command syntax as follows, because it
a) handles "pool" and "server" entries at the same time.
b) handles multiple spaces between the keyword and the value.
c) only requires two instead of three commands.
```
$servers = shell_exec('grep -E \'^pool|^server\' /etc/ntp.conf | awk -F" " \'{print $2}\';');
```
Due to the fact that it might be possible that the web server user is not allowed to read the content of `/etc/ntp.conf`, it would be a good idea to print out an error message, e.g. "/etc/ntp.conf permission denied". On my server I had to add the mentioned user to the ntp group to be allowed to read the parameters.

#### Server Configuration
OS: Linux 3.16.70
Web server: Apache2 2.4.41
Database: MariaDB 10.2.24
PHP version: 7.3.10
Nextcloud version: 17.0.0
**Serverinfo app version: 1.7.0**

#### Client Configuration
Browser: Mozilla Firefox 69.0.3
Operating system: Windows 10

tflidd:
If you propose already a fix, create a pull request. We can then review it and merge it.

radoeka:
The fix is going to work for my system, that is using the newer chrony NTP package.

```
# ls -l  /etc/ntp.conf /etc/systemd/timesyncd.conf /etc/chrony.conf
ls: cannot access '/etc/ntp.conf': No such file or directory
[...]

PR #188, opened 2020-03-23
fix for "NTP servers not displayed"

j-ed:
fix https://github.com/nextcloud/serverinfo/issues/160


j-ed:
Unfortunately I don't know how I can fix the DCO problem using the GitHub web gui. Maybe someone else can fix it for me.

tflidd:
I'm just wondering, since there are two files checked (`/etc/systemd/timesyncd.conf` and `/etc/ntp.conf`), there are systems with just one, both or none of these files. Would it be better to check if the file exists before running the command?

Issue #203, opened 2020-04-18
Patch: correct show information in FreeBSD - CPU, RAM, uptime, network (+MTU)


VVD:
### Steps to reproduce
1. Open http://my-nexcloud.my.domain/index.php/settings/admin/serverinfo
2. Look at CPU, RAM, uptime, network information

### Expected behaviour
See correct information

### Actual behaviour
No this information

### Server configuration

**Operating system**:
FreeBSD 12.1-RELEASE-p3 amd64

**Web server:**
Apache 2.4.43

**Database:**
PostgreSQL 12.2

**PHP version:**
7.4.4

**Nextcloud version:** (see Nextcloud admin page)
18.0.3.0

Don't think other information are necessary.
Patch is here: [serverinfo.diff.txt](https://github.com/nextcloud/serverinfo/files/4496997/serverinfo.diff.txt)

This patch fixes #172.

P.S. I'm not a php programmer, but this part of core is obvious. :-)

VVD:
OS name doesn't show too.
nextcloud/apps/serverinfo/templates/settings-admin.php:
```
<?php p($l->t('Operating System'));?>:
<span class="info"><?php p($_['osname']);?></span>
```
Can't find there 'osname' filled…

I can create method in DefaultOs.php.
For FreeBSD it's "uname -mrs", for Linux it can be same "uname -mrs".

VVD:
Disk size show with default block size 512 bytes instead of 1KB.
This patch work fine for both FreeBSD and Linux:
nextcloud/apps/serverinfo/lib/OperatingSystems/DefaultOs.php:
```
-                $data  = shell_exec('df -TP');
+                $data  = shell_exec('df -TPk');
```

Swap size show with default block size 512 bytes too:
```
$ BLOCKSIZE= swapinfo 
Device          512-blocks     Used    Avail Capacity
/dev/da1           8388608   214560  8174048     3%
$ BLOCKSIZE=K swapinfo 
Device          1K-blocks     Used    Avail Capacity
/dev/da1          4194304   107280  4087024     3%
$ swapinfo -k 
Device          1K-blocks     Used    Avail Capacity
/dev/da1          4194304   107280  4087024     3%
$ swapinfo -m
Device          1M-blocks     Used    Avail Capacity
/dev/da1             4096      103     3992     3%
$ swapinfo -g
Device          1G-blocks     Used    Avail Capacity
/dev/da1                4        0        3     3%
```
Nextcloud show this information:
[...]

PR #209, opened 2020-06-06
Extend ServerInfo support to FreeBSD

constrict:
FreeBSD does not use /proc/ and values are set to unknown/-1


kesselb:
https://github.com/nextcloud/serverinfo/issues/203#issuecomment-616232757 how to do it properly.

constrict:
Thanks. Will work on doing this correctly.

constrict:
Memory, uptime, CPU, network gateway, network interfaces are now available.

constrict:
I am finished with this now. I made the changes needed to pass

Issue #172, opened 2020-02-07
A non-numeric value encountered at DefaultOs.php#52


pixel8383:
When browsing the /serverinfo section of my admin panel, I get on the error log:

`A non-numeric value encountered at /usr/local/www/nextcloud/apps/serverinfo/lib/OperatingSystems/DefaultOs.php#52`

The page itself seems ok, it's displayed as expected anyway.

tflidd:
Can you give more information about your system?
If you have terminal access, could you run this command:
`cat /proc/meminfo | grep MemTotal`

It seems that parsing of this output goes wrong.

pixel8383:
It's a freeBSD machine, so I can use `sysctl hw.realmem`
The answer is:
`hw.realmem: 25769803776`

Soygen:
I'm seeing this same error in my log. I'm also running FreeBSD(FreeNAS system with NextCloud running in a jail). Have you figured out how to resolve this?

pvuchetich:
I also see the same issue using the FreeNAS 11.2-U7 Nextcloud plugin. The technical issue is DefaultOs.php parses a Linux-specific file to extract system information. In FreeNAS (FreeBSD 11.2 OS) jail, the "/proc" directory is empty, so the fix would be to use an OS-specific command.  FreeBSD isn't on the list of supported platforms, so there are two ways to resolve this:

See if there are FreeNAS-specific workarounds (community forum for FreeNAS, with a specific group for the Nextcloud plugin) (I did not find one, so I posted a new message)

https://www.ixsystems.com/community/threads/new-install-of-nextcloud-error-message-with-a-non-numeric-value-encountered-at-defaultos-php-52.83329/

The general forum for the FreeNAS Nextcould plugin:
https://www.ixsystems.com/community/forums/Nextcloud/

Or...
Even though it is not an officially supported OS, would it be worth making a pull request with changes to the server information page to check the OS, and use FreeBSD-specific commands? It might take me a bit to figure this out, as I haven't used Nextcloud or looked at its code before today.

pixel8383:
Another easy fix? This one: check the OS version and if it's not a valid one, just avoid to extract informations.
I know it would be better to build FreeBSD specific queries but if the OS is listed as non supported, this could just fix the issue.

VVD:
@pixel8383, @pvuchetich, @Soygen, plz, check my patch at #203.

kesselb:
Fixed by #210 for Nextcloud 20.

PR #210, opened 2020-06-07
Extend ServerInfo to Support FreeBSD

constrict:
FreeBSD doesn't use /proc/ to store system information, so multiple functions needed to be modified to support FreeBSD,

Modified Os.php to include new Class FreeBSD added check for when FreeBSD is the OS.

Created FreeBSD.php based on DefaultOS.php

Functions modified from DefaultOS.php
- Added constructor to use phpIni
- **getMemory** - use sysctl - mirror logic from SystemStatistics.php
- **getCPUName** - use sysctl -n hw.model
- **getUptime** - return kern.bootime subtracted from current time
- **getTimeServers** - remove /etc/systemd/timesyncd.conf as does not exist in FreeBSD
- **getNetworkInfo** - add gateway using netstat
- **getNetworkInterfaces** - modified to use ifconfig instead of /sys/class/net/*
- **readContent** - removed as now unused

### DIFF
--- a/lib/OperatingSystems/FreeBSD.php
+++ b/lib/OperatingSystems/FreeBSD.php
@@ -102,9 +102,10 @@ class FreeBSD {
 	 */
 	public function getUptime(): int {
 		try {
-			$uptime = $this->executeCommand('/sbin/sysctl -n kern.boottime | tr -d \',\' | cut -d \' \' -f4');
+			$shell_boot = $this->executeCommand('/sbin/sysctl -n kern.boottime');
+			preg_match("/[\d]+/", $shell_boot, $boottime);
 			$time = $this->executeCommand('date +%s');
-			$uptimeInSeconds = (int)$uptime - (int)$time;
+			$uptimeInSeconds = (int)$time - (int)$boottime[0];
 		} catch (\RuntimeException $e) {
 			return $uptimeInSeconds;
 		}
@@ -116,11 +117,13 @@ class FreeBSD {
 	 */
 	public function getTimeServers() {
 		try {
-			$servers = $this->executeCommand('cat /etc/ntp.conf 2>/dev/null |grep  \'^pool\' | cut -f 2 -d " "');
+			$servers = $this->executeCommand('cat /etc/ntp.conf 2>/dev/null');
+			preg_match_all("/(?<=^pool ).\S*/m", $servers, $matches);
+			$allservers = implode(' ', $matches[0]);
 		} catch (\RuntimeException $e) {
 			return $servers;
 		}
-		return $servers;
+		return $allservers;
 	}
 
 	/**
@@ -131,10 +134,13 @@ class FreeBSD {
 		$result['hostname'] = \gethostname();
 		
 		try {
-			$dns = $this->executeCommand('cat /etc/resolv.conf |grep -i \'^nameserver\'|head -n1|cut -d \' \' -f2');
-			$result['dns'] = $dns;
-			$gw = $this->executeCommand('netstat -rn | grep default | cut -d \' \' -f13');
-			$result['gateway'] = $gw;
+			$dns = $this->executeCommand('cat /etc/resolv.conf 2>/dev/null');
+			preg_match_all("/(?<=^nameserver ).\S*/m", $dns, $matches);
+			$alldns = implode(' ', $matches[0]);
+			$result['dns'] = $alldns;
+			$netstat = $this->executeCommand('netstat -rn');
+			preg_match("/(?<=^default).*\b\d/m", $netstat, $gw);
+			$result['gateway'] = $gw[0];
 		} catch (\RuntimeException $e) {
 			return $result;
 		}
@@ -147,27 +153,36 @@ class FreeBSD {
 	public function getNetworkInterfaces() {
 		$result = [];
 		
-		$interfaces = $this->executeCommand('/sbin/ifconfig -a | cut -d$\'\t\' -f1 | cut -d \':\' -f1 | grep -v -e \'^$\'');
-		$interfaces = preg_split('/\s+/', trim($interfaces));
-
-		foreach ($interfaces as $interface) {
+		$ifconfig = $this->executeCommand('/sbin/ifconfig -a');
+		preg_match_all("/^(?<=(?!\t)).*(?=:)/m", $ifconfig, $interfaces);
+		
+		foreach ($interfaces[0] as $interface) {
 			$iface              = [];
 			$iface['interface'] = $interface;
-			$iface['ipv4']      = $this->executeCommand('/sbin/ifconfig ' . $iface['interface'] . ' | grep \'inet \' | cut -f2 -d$\'\t\' | cut -f 2 -d \' \'');
-			$iface['ipv6']      = $this->executeCommand('/sbin/ifconfig ' . $iface['interface'] . ' | grep inet6 | cut -f2 -d$\'\t\' | cut -f 2 -d \' \' | cut -f1 -d \'%\'');
+			$intface            = $this->executeCommand('/sbin/ifconfig ' . $iface['interface']);
+			preg_match_all("/(?<=inet ).\S*/m", $intface, $ipv4);
+			preg_match_all("/(?<=inet6 ).*(?=%)/m", $intface, $ipv6);
+			$iface['ipv4']      = implode(' ', $ipv4[0]);
+			$iface['ipv6']      = implode(' ', $ipv6[0]);
+
 			if ($iface['interface'] !== 'lo0') {
-				$iface['mac']       = $this->executeCommand('/sbin/ifconfig ' . $iface['interface'] . ' | grep ether | cut -f2 -d$\'\t\' | cut -f2 -d \' \'');
-				$iface['status'] = $this->executeCommand('/sbin/ifconfig ' . $iface['interface'] . ' | grep status | cut -f2 -d$\'\t\' | cut -f2 -d \' \'');
-				$iface['speed']  = $this->executeCommand('/sbin/ifconfig ' . $iface['interface'] . ' | grep media | cut -d \' \' -f3 | cut -f1 -d \'b\'');
+				preg_match_all("/(?<=ether ).*/m", $intface, $mac);
+				preg_match("/(?<=status: ).*/m", $intface, $status);
+				preg_match("/(?<=\().*(?=b)/m", $intface, $speed);
+				preg_match("/(?<=\<).*(?=-)/m", $intface, $duplex);
+				
+				$iface['mac'] = implode(' ', $mac[0]);
+				$iface['status'] = $status[0];
+				$iface['speed']  = $speed[0];
+				
 				if ($iface['speed'] !== '') {
 					$iface['speed'] = $iface['speed'];
 				} else {
 					$iface['speed'] = 'unknown';
 				}
 
-				$duplex = $this->executeCommand('/sbin/ifconfig ' . $iface['interface'] . ' | grep media | cut -d \'<\' -f2 | cut -d \'-\' -f1');
-				if ($duplex !== '') {
-					$iface['duplex'] = 'Duplex: ' . $duplex;
+				if ($duplex[0] !== '') {
+					$iface['duplex'] = 'Duplex: ' . $duplex[0];
 				} else {
 					$iface['duplex'] = '';
 				}
@@ -227,7 +242,7 @@ class FreeBSD {
 	}
 
 	protected function executeCommand(string $command): string {
-		$output = @shell_exec($command);
+		$output = @shell_exec(escapeshellcmd($command));
 		if ($output === null || $output === '') {
 			throw new \RuntimeException('No output for command: "' . $command . '"');
 		}
--- a/lib/OperatingSystems/FreeBSD.php
+++ b/lib/OperatingSystems/FreeBSD.php
@@ -161,7 +161,7 @@ class FreeBSD {
 			$iface['interface'] = $interface;
 			$intface            = $this->executeCommand('/sbin/ifconfig ' . $iface['interface']);
 			preg_match_all("/(?<=inet ).\S*/m", $intface, $ipv4);
-			preg_match_all("/(?<=inet6 ).*(?=%)/m", $intface, $ipv6);
+			preg_match_all("/(?<=inet6 )((.*(?=%))|(.\S*))/m", $intface, $ipv6);
 			$iface['ipv4']      = implode(' ', $ipv4[0]);
 			$iface['ipv6']      = implode(' ', $ipv6[0]);
 
--- a/lib/OperatingSystems/FreeBSD.php
+++ b/lib/OperatingSystems/FreeBSD.php
@@ -47,25 +47,24 @@ class FreeBSD {
 
 		try {
 			$swapinfo = $this->executeCommand('/usr/sbin/swapinfo');
+ 			$line = preg_split("/[\s]+/", $swapinfo);
+
+			if (count($line) > 3) {
+				$data['SwapTotal'] = (int)$line[3];
+				$data['SwapFree'] = $data['SwapTotal'] - (int)$line[2];
+			}
+			
+			$meminfo = $this->executeCommand('/sbin/sysctl -n hw.physmem hw.pagesize vm.stats.vm.v_inactive_count vm.stats.vm.v_cache_count vm.stats.vm.v_free_count');
+
+			$line = preg_split('/\s+/', trim($meminfo));
+			if (count($line) > 4) {
+				$data['MemTotal'] = (int)$line[0];
+				$data['MemAvailable'] = (int)$line[1] * ((int)$line[2] + (int)$line[3] + (int)$line[4]);
+			}
 		} catch (\RuntimeException $e) {
 			return $data;
 		}
 
-		$line = preg_split("/[\s]+/", $swapinfo);
-		if (count($line) > 3) {
-			$data['SwapTotal'] = (int)$line[3];
-			$data['SwapFree'] = $data['SwapTotal'] - (int)$line[2];
-		}
-
-		try {
-			$return = $this->executeCommand('/sbin/sysctl -n hw.physmem hw.pagesize vm.stats.vm.v_inactive_count vm.stats.vm.v_cache_count vm.stats.vm.v_free_count');
-			$return = preg_split('/\s+/', trim($return));
-
-			$data['MemTotal'] = (int)$return[0];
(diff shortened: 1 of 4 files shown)


lib/OperatingSystems/FreeBSD.php
```diff
@@ -0,0 +1,239 @@
+<?php
+/**
+ * @author Frank Karlitschek <frank@nextcloud.com>
```
kesselb, line 3: Enter your name here.

lib/OperatingSystems/FreeBSD.php
```diff
@@ -0,0 +1,239 @@
+<?php
```
kesselb, line 1:
```suggestion
<?php
declare(strict_types=1);
```

lib/OperatingSystems/FreeBSD.php
```diff
@@ -0,0 +1,239 @@
...
+	protected $phpIni;
+	
+	/**
+	 * FreeBSD constructor.
+	 *
+	 * @param IniGetWrapper $phpIni
+	 * @throws \Exception
+	 */
+	public function __construct(IniGetWrapper $phpIni) {
+		$this->phpIni = $phpIni;
```
kesselb, line 42: We don't need the IniGetWrapper. Please remove it again.

lib/OperatingSystems/FreeBSD.php
```diff
@@ -0,0 +1,239 @@
...
+	}
+	
+	/**
+	 * @return bool
+	 */
+	public function supported(): bool {
+		return true;
```
kesselb, line 49:
```suggestion
		return false;
```

because FreeBSD is not supported.

lib/OperatingSystems/FreeBSD.php
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	/**
+	 * Checks if a function is available. Borrowed from
+	 * https://github.com/nextcloud/server/blob/2e36069e24406455ad3f3998aa25e2a949d1402a/lib/private/legacy/helper.php#L475
+	 *
+	 * @param string $function_name
+	 * @return bool
+	 */
+	public function is_function_enabled($function_name) {
```
kesselb, line 230: Remove this method.

lib/OperatingSystems/FreeBSD.php
```diff
@@ -0,0 +1,239 @@
...
+	/**
+	 * Get memory will return a list key => value where all values are in bytes.
+	 * [MemTotal => 0, MemFree => 0, MemAvailable => 0, SwapTotal => 0, SwapFree => 0].
+	 *
+	 * @return array
+	 */
+	public function getMemory(): array {
+		$data = ['MemTotal' => -1, 'MemFree' => -1, 'MemAvailable' => -1, 'SwapTotal' => -1, 'SwapFree' => -1];
+ 
+		$swapinfo = shell_exec('/usr/sbin/swapinfo');
```
kesselb, line 61:
```suggestion
		try {
			$swapinfo = $this->executeCommand('/usr/sbin/swapinfo');
		} catch (\RuntimeException $e) {
			return $data;
		}
```

executeCommand is a method I added to DefaultOS as alternative to `is_function_enabled`. Please use it here for every execution of shell_exec and exec. It's actually a wrapper to makes it possible to mock/overwrite the function calls and write tests for the parsing logic.

lib/OperatingSystems/FreeBSD.php
constrict: Done

lib/OperatingSystems/FreeBSD.php
constrict: Done

lib/OperatingSystems/FreeBSD.php
constrict: Done

lib/OperatingSystems/FreeBSD.php
constrict: Done

lib/OperatingSystems/FreeBSD.php
constrict: Done

lib/OperatingSystems/FreeBSD.php
constrict: Done. Had to remove escapeshellcmd form executeCommand because of issues with piped commands

kesselb: Hmm. I would prefer to keep escapeshellcmd. What is done with pipes can also done with PHP.

constrict: It might take a bit but I think I can do this.

kesselb: Thanks. If you need help paste the raw command output here and I can have a look how to parse it.

kesselb: serverinfo app is always released with Nextcloud hence we have some time until feature freeze for 20.

constrict: Turns out regular expressions are not as hard as I thought it would. I have replaced all piped shell commands with the base commands and cleaned with regex.

lib/OperatingSystems/FreeBSD.php
```diff
@@ -0,0 +1,251 @@
...
+	/**
+	 * Get name of the processor
+	 *
+	 * @return string
+	 */
+	public function getCPUName(): string {
+		try {
+			$data = $this->executeCommand('/sbin/sysctl -n hw.model');
+		} catch (\RuntimeException $e) {
+			return $data;
```
kesselb, line 81: If a runtime exception occurs $data might be undefined. Please add a default like: https://github.com/nextcloud/serverinfo/blob/c571f369ad2e112c1aab70ad436a940e8ad6ca77/lib/OperatingSystems/DefaultOs.php#L80

_(fyi: I will probably move those defaults to Os.php for DefaultOs and FreeBSD in a follow up pr)_

lib/OperatingSystems/FreeBSD.php
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getTime() {
+		try {
+			$uptime = $this->executeCommand('date');
+		} catch (\RuntimeException $e) {
+			return $uptime;
```
kesselb, line 93: Same here. On runtime exception $update might be undefined. Please add a default. A empty string should do it.

lib/OperatingSystems/FreeBSD.php
```diff
@@ -0,0 +1,251 @@
...
+	 * @return int
+	 */
+	public function getUptime(): int {
+		try {
+			$shell_boot = $this->executeCommand('/sbin/sysctl -n kern.boottime');
+			preg_match("/[\d]+/", $shell_boot, $boottime);
+			$time = $this->executeCommand('date +%s');
+			$uptimeInSeconds = (int)$time - (int)$boottime[0];
+		} catch (\RuntimeException $e) {
+			return $uptimeInSeconds;
```
kesselb, line 110: Please define the variable $uptimeInSeconds with a default value of -1 before the try block.

lib/OperatingSystems/FreeBSD.php
```diff
@@ -0,0 +1,251 @@
...
+			$data['SwapTotal'] = (int)$line[3];
+			$data['SwapFree'] = $data['SwapTotal'] - (int)$line[2];
+		}
+
+		try {
+			$return = $this->executeCommand('/sbin/sysctl -n hw.physmem hw.pagesize vm.stats.vm.v_inactive_count vm.stats.vm.v_cache_count vm.stats.vm.v_free_count');
+			$return = preg_split('/\s+/', trim($return));
+
+			$data['MemTotal'] = (int)$return[0];
+			$data['MemAvailable'] = (int)$return[1] * ((int)$return[2] + (int)$return[3] + (int)$return[4]);
```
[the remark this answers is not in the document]
constrict: Done

lib/OperatingSystems/FreeBSD.php
constrict: Done

lib/OperatingSystems/FreeBSD.php
constrict: Done

lib/OperatingSystems/FreeBSD.php
constrict: Done

lib/OperatingSystems/FreeBSD.php
```diff
@@ -0,0 +1,258 @@
...
+			return $data;
+		}
+		return $data;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getTime() {
+		$time = " ";
```
kesselb, line 91:
```suggestion
		$time = '';
```

kesselb:
Mind to share some of the command outputs here (like https://github.com/nextcloud/serverinfo/tree/master/tests/data)? I can add some tests later. Just post it as comments here. No need to add it as files yet ;) Thanks :+1:

constrict:
Sample Outputs
**/usr/sbin/swapinfo**
```
Device          1K-blocks     Used    Avail Capacity
/dev/ada0p3       3744300        0  3744300     0%
```
**/sbin/sysctl -n hw.physmem hw.pagesize vm.stats.vm.v_inactive_count vm.stats.vm.v_cache_count vm.stats.vm.v_free_count**

constrict:
7. **cat /etc/ntp.conf 2>/dev/null**
```
#
# $FreeBSD: releng/12.1/usr.sbin/ntp/ntpd/ntp.conf 337649 2018-08-11 17:42:42Z brd $
#
# Default NTP servers for the FreeBSD operating system.
#
# Don't forget to enable ntpd in /etc/rc.conf with:
# ntpd_enable="YES"
#
# The driftfile is by default /var/db/ntpd.drift, check
# /etc/defaults/rc.conf on how to change the location.
#

#
# Set the target and limit for adding servers configured via pool statements
# or discovered dynamically via mechanisms such as broadcast and manycast.
# Ntpd automatically adds maxclock-1 servers from configured pools, and may
# add as many as maxclock*2 if necessary to ensure that at least minclock
# servers are providing good consistant time.
#
tos minclock 3 maxclock 6

#
# The following pool statement will give you a random set of NTP servers
# geographically close to you.  A single pool statement adds multiple
# servers from the pool, according to the tos minclock/maxclock targets.
# See http://www.pool.ntp.org/ for details.  Note, pool.ntp.org encourages
# users with a static IP and good upstream NTP servers to add a server
# to the pool. See http://www.pool.ntp.org/join.html if you are interested.
#
# The option `iburst' is used for faster initial synchronization.
#
pool 0.freebsd.pool.ntp.org iburst

#
# If you want to pick yourself which country's public NTP server
# you want to sync against, comment out the above pool, uncomment
# the next one, and replace CC with the country's abbreviation.
# Make sure that the hostname resolves to a proper IP address!
#
# pool 0.CC.pool.ntp.org iburst

#
# To configure a specific server, such as an organization-wide local
# server, add lines similar to the following.  One or more specific
# servers can be configured in addition to, or instead of, any server
# pools specified above.  When both are configured, ntpd first adds all
# the specific servers, then adds servers from the pool until the tos
# minclock/maxclock targets are met.
#
#server time.my-internal.org iburst

server 0.north-america.pool.ntp.org
server 1.north-america.pool.ntp.org
server 2.north-america.pool.ntp.org
server 3.north-america.pool.ntp.org

driftfile /var/db/ntp.drift

#
# Security:
#
# By default, only allow time queries and block all other requests
# from unauthenticated clients.
#
# The "restrict source" line allows peers to be mobilized when added by
# ntpd from a pool, but does not enable mobilizing a new peer association
# by other dynamic means (broadcast, manycast, ntpq commands, etc).
#
# See http://support.ntp.org/bin/view/Support/AccessRestrictions
# for more information.
#
restrict default limited kod nomodify notrap noquery nopeer
restrict source  limited kod nomodify notrap noquery

#
# Alternatively, the following rules would block all unauthorized access.
#
#restrict default ignore
#
# In this case, all remote NTP time servers also need to be explicitly
# allowed or they would not be able to exchange time information with
# this server.
#
# Please note that this example doesn't work for the servers in
# the pool.ntp.org domain since they return multiple A records.
#
#restrict 0.pool.ntp.org nomodify nopeer noquery notrap
#restrict 1.pool.ntp.org nomodify nopeer noquery notrap
#restrict 2.pool.ntp.org nomodify nopeer noquery notrap
#
# The following settings allow unrestricted access from the localhost
restrict 127.0.0.1
restrict ::1

#
# If a server loses sync with all upstream servers, NTP clients
# no longer follow that server. The local clock can be configured
# to provide a time source when this happens, but it should usually
# be configured on just one server on a network. For more details see
# http://support.ntp.org/bin/view/Support/UndisciplinedLocalClock
# The use of Orphan Mode may be preferable.
#
#server 127.127.1.0
#fudge 127.127.1.0 stratum 10

# See http://support.ntp.org/bin/view/Support/ConfiguringNTP#Section_6.14.
# for documentation regarding leapfile. Updates to the file can be obtained
# from ftp://time.nist.gov/pub/ or ftp://tycho.usno.navy.mil/pub/ntp/.
# Use either leapfile in /etc/ntp or periodically updated leapfile in /var/db.
#leapfile "/etc/ntp/leap-seconds"
leapfile "/var/db/ntpd.leap-seconds.list"
```

constrict:
8. **cat /etc/resolv.conf 2>/dev/null**
```
# Generated by resolvconf
nameserver 10.0.0.1
nameserver 2001:558:feed::1
nameserver 2001:558:feed::2
```

kesselb:
Thanks and don't forget to reopen the pr and merge :)

constrict:
9. **netstat -rn**
```
Routing tables

Internet:
Destination        Gateway            Flags     Netif Expire
default            10.0.0.1           UGS     epair0b
10.0.0.0/24        link#2             U       epair0b
10.0.0.185         link#2             UHS         lo0
127.0.0.1          link#1             UH          lo0
```

constrict:
10. **/sbin/ifconfig -a**
```
**lo0: flags=8049<UP,LOOPBACK,RUNNING,MULTICAST> metric 0 mtu 16384
        options=680003<RXCSUM,TXCSUM,LINKSTATE,RXCSUM_IPV6,TXCSUM_IPV6>
        inet6 ::1 prefixlen 128
        inet6 0000::1%lo0 prefixlen 64 scopeid 0x1
        inet 127.0.0.1 netmask 0xff000000
        groups: lo
        nd6 options=21<PERFORMNUD,AUTO_LINKLOCAL>
epair0b: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 1500
        options=8<VLAN_MTU>
        ether 00:00:00:00:00:00
        hwaddr 00:00:00:00:00:00
        inet 10.0.0.2 netmask 0xffffff00 broadcast 10.0.0.255
        inet6 0000::00:0000:0000:0000%epair0b prefixlen 64 scopeid 0x2
        groups: epair
        media: Ethernet 10Gbase-T (10Gbase-T <full-duplex>)
        status: active
        nd6 options=21<PERFORMNUD,AUTO_LINKLOCAL>**
```

constrict:
> Thanks and don't forget to reopen the pr and merge :)

Woops closed by accident. I cannot merge it myself.

constrict:
#172 #203 Should resolve both

kesselb:
Thanks :+1:

VVD:
1. "df -TP" use 512 bytes blocks size - correct command is "df -TPk".
2. No MTU - check 1st line:
```
ixl0: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 9198
        options=e507bb<RXCSUM,TXCSUM,VLAN_MTU,VLAN_HWTAGGING,JUMBO_MTU,VLAN_HWCSUM,TSO4,TSO6,LRO,VLAN_HWFILTER,VLAN_HWTSO,RXCSUM_IPV6,TXCSUM_IPV6>
   
     ether XX:XX:XX:XX:XX:XX
        media: Ethernet autoselect (10Gbase-T <full-duplex>)
        status: active
        nd6 options=29<PERFORMNUD,IFDISABLED,AUTO_LINKLOCAL>
```
3. No CPU cores count: "sysctl -n kern.smp.cpus".
4. Total RAM is hw.realmem, but not hw.physmem.

constrict:
Issues 1, 3 and 4 are resolved in #215 
Adding mtu is an enhancement and needs to be added to the default OS first. I have no interest in mtu so I suggest you create a pull yourself @VVD

constrict:
For the record the change is simple in freebsd

```
if ($iface['interface'] !== 'lo0') {
	preg_match_all("/(?<=ether ).*/m", $intface, $mac);
	preg_match_all("/(?<=mtu ).*/m", $intface, $mtu);
	preg_match("/(?<=status: ).*/m", $intface, $status);
	preg_match("/(?<=\().*(?=b)/m", $intface, $speed);
	preg_match("/(?<=\<).*(?=-)/m", $intface, $duplex);

	$iface['mac'] = implode(' ', $mac[0]);
	$iface['mtu'] = $mtu[0];
	$iface['status'] = $status[0];
	$iface['speed']  = $speed[0]
```

VVD:
5. swapinfo need -k same as df: "/usr/sbin/swapinfo -k".

@constrict, I wrote patch 2 months ago: #203.

constrict:
Does not seem to be an issue on my box but I have added -k to swapinfo, so that it is always shown in 1k-blocks.

constrict:
> Does not seem to be an issue on my box but I have added -k to swapinfo, so that it is always shown in 1k-blocks.

Issue is actually in SystemStatistics.php. The swap return is unused in FreeBSD.php.

EliranHayun:
Hi all,

I have FreeNAS (FreeBSD) system running Nextcloud 18.0.6
Is there anything I can help with?

Thanks,
Eliran

Schueni1:
I'm still getting an '_Internal Server Error_' when I try to access settings/admin/serverinfo on FreeBSD 12.1 with NC20.
Shouldn't this already work?

lazeraman:
# #245 Should Fix this

constrict:
> I'm still getting an '_Internal Server Error_' when I try to access settings/admin/serverinfo on FreeBSD 12.1 with NC20.
> Shouldn't this already work?

See #245

PaulArgoud:
> > I'm still getting an '_Internal Server Error_' when I try to access settings/admin/serverinfo on FreeBSD 12.1 with NC20.
> > Shouldn't this already work?
> 
> See #245 and #251

Same problem for me, with FreeBSD 12 and NC20...

PR #212, opened 2020-06-16
Add status information about the opcache and APCu instances

### LINKED ISSUE #196: Extension APC / APCu
As an extension a Informationblock about APC / APCu would be practical.

for example: 

https://raw.githubusercontent.com/krakjoe/apcu/master/apc.php

it generates a detailed informationpage, that could served as a template.

### LINKED ISSUE #197: Extension OPcache
As an extension a Informationblock about the PHP opcache would be practical.

an example is given in an discussion entry:

https://help.nextcloud.com/t/http-error-500-when-trying-to-access-nextcloud/33646/3

it generates a detailed informationpage, that could served as a template.

Starkstromkonsument:
closes #196, closes #197

### DIFF
--- a/lib/PhpStatistics.php
+++ b/lib/PhpStatistics.php
@@ -72,23 +72,21 @@ class PhpStatistics {
 	 *
 	 * @return array with an array of state information about the cache instance
 	 */
-	protected function getOPcacheStatus() {
+	protected function getOPcacheStatus(): array {
 		// Test if the OPcache module is installed
 		if (!extension_loaded('Zend OPcache')) {
-			// module not loaded
-			// returning back zeroes to prevent any errors on JS side.
-			return 'N/A';
+			// module not loaded, returning back empty array to prevent any errors on JS side.
+			return [];
 		}
 
 		// get status information about the cache
 		$status = opcache_get_status(false);
 
-		if (is_array($status)) {
+		if ($status !== false) {
 			return $status;
 		} else {
-			// no array
-			// returning back zeroes to prevent any errors on JS side.
-			return 'N/A';
+			// no array, returning back empty array to prevent any errors on JS side.
+			return [];
 		}
 	}
 
@@ -97,12 +95,11 @@ class PhpStatistics {
 	 *
 	 * @return array with an array of state information about the cache instance
 	 */
-	protected function getAPCuStatus() {
+	protected function getAPCuStatus(): array {
 		// Test if the APCu module is installed
 		if (!extension_loaded('apcu')) {
-			// module not loaded
-			// returning back zeroes to prevent any errors on JS side.
-			return 'N/A';
+			// module not loaded, returning back empty array to prevent any errors on JS side.
+			return [];
 		}
 
 		// get cached information from APCu data store
@@ -112,15 +109,13 @@ class PhpStatistics {
 		$smaInfo = apcu_sma_info(true);
 
 		if (!is_array($cacheInfo)) {
-			// no array
-			// returning back zeroes to prevent any errors on JS side.
-			$cacheInfo = 'N/A';
+			// no array, returning back empty array to prevent any errors on JS side.
+			$cacheInfo = [];
 		}
 
 		if (!is_array($smaInfo)) {
-			// no array
-			// returning back zeroes to prevent any errors on JS side.
-			$smaInfo = 'N/A';
+			// no array, returning back empty array to prevent any errors on JS side.
+			$smaInfo = [];
 		}
 
 		// return the array
--- a/lib/PhpStatistics.php
+++ b/lib/PhpStatistics.php
@@ -106,27 +106,27 @@ class PhpStatistics {
 		}
 
 		// get cached information from APCu data store
-		$cache_info = apcu_cache_info(true);
+		$cacheInfo = apcu_cache_info(true);
 
 		// get APCu Shared Memory Allocation information
-		$sma_info = apcu_sma_info(true);
+		$smaInfo = apcu_sma_info(true);
 
-		if (!is_array($cache_info)) {
+		if (!is_array($cacheInfo)) {
 			// no array
 			// returning back zeroes to prevent any errors on JS side.
-			$cache_info = 'N/A';
+			$cacheInfo = 'N/A';
 		}
 
-		if (!is_array($sma_info)) {
+		if (!is_array($smaInfo)) {
 			// no array
 			// returning back zeroes to prevent any errors on JS side.
-			$sma_info = 'N/A';
+			$smaInfo = 'N/A';
 		}
 
 		// return the array
 		return [
-			'cache' => $cache_info,
-			'sma' => $sma_info,
+			'cache' => $cacheInfo,
+			'sma' => $smaInfo,
 		];
 	}
 }
--- a/lib/PhpStatistics.php
+++ b/lib/PhpStatistics.php
@@ -48,6 +48,7 @@ class PhpStatistics {
 			'max_execution_time' => $this->phpIni->getNumeric('max_execution_time'),
 			'upload_max_filesize' => $this->phpIni->getBytes('upload_max_filesize'),
 			'opcache' => $this->getOPcacheStatus(),
+			'apcu' => $this->getAPCuStatus(),
 		];
 	}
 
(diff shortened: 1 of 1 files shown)


lib/PhpStatistics.php
```diff
@@ -64,4 +66,67 @@ protected function cleanVersion($version) {
 		}
 		return $version;
 	}
+
+	/**
+	 * Get status information about the cache from the OPcache extension
+	 *
+	 * @return array with an array of state information about the cache instance
+	 */
+	protected function getOPcacheStatus() {
```
kesselb, line 75: I would prefer to return a empty array if no information is found. That makes it possible add the return type.

kesselb, line 75:
```suggestion
	protected function getOPcacheStatus(): array {
```

lib/PhpStatistics.php
```diff
@@ -64,4 +66,67 @@ protected function cleanVersion($version) {
...
+		// Test if the OPcache module is installed
+		if (!extension_loaded('Zend OPcache')) {
+			// module not loaded
+			// returning back zeroes to prevent any errors on JS side.
+			return 'N/A';
+		}
+
+		// get status information about the cache
+		$status = opcache_get_status(false);
```
kesselb, line 84: If plesk is used this method is disabled by default.

lib/PhpStatistics.php
```diff
@@ -64,4 +66,67 @@ protected function cleanVersion($version) {
...
+	 * Get status information about the cache from the OPcache extension
+	 *
+	 * @return array with an array of state information about the cache instance
+	 */
+	protected function getOPcacheStatus() {
+		// Test if the OPcache module is installed
+		if (!extension_loaded('Zend OPcache')) {
+			// module not loaded
+			// returning back zeroes to prevent any errors on JS side.
+			return 'N/A';
```
kesselb, line 80:
```suggestion
			return [];
```

lib/PhpStatistics.php
```diff
@@ -64,4 +66,67 @@ protected function cleanVersion($version) {
...
+
+		// get status information about the cache
+		$status = opcache_get_status(false);
+
+		if (is_array($status)) {
+			return $status;
+		} else {
+			// no array
+			// returning back zeroes to prevent any errors on JS side.
+			return 'N/A';
```
kesselb, line 91:
```suggestion
			return [];
```

lib/PhpStatistics.php
```diff
@@ -64,4 +66,67 @@ protected function cleanVersion($version) {
...
+		if (!extension_loaded('Zend OPcache')) {
+			// module not loaded
+			// returning back zeroes to prevent any errors on JS side.
+			return 'N/A';
+		}
+
+		// get status information about the cache
+		$status = opcache_get_status(false);
+
+		if (is_array($status)) {
```
kesselb, line 86:
```suggestion
		if ($status !== false) {
```

lib/PhpStatistics.php
```diff
@@ -64,4 +66,67 @@ protected function cleanVersion($version) {
...
+	protected function getAPCuStatus() {
+		// Test if the APCu module is installed
+		if (!extension_loaded('apcu')) {
+			// module not loaded
+			// returning back zeroes to prevent any errors on JS side.
+			return 'N/A';
+		}
+
+		// get cached information from APCu data store
+		$cache_info = apcu_cache_info(true);
```
kesselb, line 109:
```suggestion
		$cacheInfo = apcu_cache_info(true);
```

lib/PhpStatistics.php
```diff
@@ -64,4 +66,67 @@ protected function cleanVersion($version) {
...
+
+		// get APCu Shared Memory Allocation information
+		$sma_info = apcu_sma_info(true);
```
kesselb, line 112:
```suggestion
		$smaInfo = apcu_sma_info(true);
```

lib/PhpStatistics.php
```diff
@@ -64,4 +66,67 @@ protected function cleanVersion($version) {
...
+
+		if (!is_array($cache_info)) {
```
kesselb, line 114:
```suggestion
		if (!is_array($cacheInfo)) {
```

lib/PhpStatistics.php
```diff
@@ -64,4 +66,67 @@ protected function cleanVersion($version) {
...
+			// no array
+			// returning back zeroes to prevent any errors on JS side.
+			$cache_info = 'N/A';
```
kesselb, line 117:
```suggestion
			$cacheInfo = 'N/A';
```

lib/PhpStatistics.php
```diff
@@ -64,4 +66,67 @@ protected function cleanVersion($version) {
...
+		}
+
+		if (!is_array($sma_info)) {
```
kesselb, line 120:
```suggestion
		if (!is_array($smaInfo)) {
```

lib/PhpStatistics.php
```diff
@@ -64,4 +66,67 @@ protected function cleanVersion($version) {
...
+			// no array
+			// returning back zeroes to prevent any errors on JS side.
+			$sma_info = 'N/A';
```
kesselb, line 123:
```suggestion
			$smaInfo = 'N/A';
```

lib/PhpStatistics.php
```diff
@@ -64,4 +66,67 @@ protected function cleanVersion($version) {
...
+			return 'N/A';
+		}
+	}
+
+	/**
+	 * Get status information about the cache from the APCu extension
+	 *
+	 * @return array with an array of state information about the cache instance
+	 */
+	protected function getAPCuStatus() {
```
kesselb, line 100:
```suggestion
	protected function getAPCuStatus(): array {
```

lib/PhpStatistics.php
```diff
@@ -64,4 +66,67 @@ protected function cleanVersion($version) {
...
+
+		if (!is_array($sma_info)) {
+			// no array
+			// returning back zeroes to prevent any errors on JS side.
+			$sma_info = 'N/A';
+		}
+
+		// return the array
+		return [
+			'cache' => $cache_info,
```
kesselb, line 128:
```suggestion
			'cache' => $cacheInfo,
```

lib/PhpStatistics.php
```diff
@@ -64,4 +66,67 @@ protected function cleanVersion($version) {
...
+			'sma' => $sma_info,
```
kesselb, line 129:
```suggestion
			'sma' => $smaInfo,
```

kesselb:
Do we need to expose all the information? I don't see what this is good for ;)

lib/PhpStatistics.php
Starkstromkonsument: I could add a check for this by using https://www.php.net/manual/en/function.function-exists.php. What do you think? I guess we can't do anything about this plesk issue. Most people confronted with this problem will probably not have the permissions to change this.

lib/PhpStatistics.php
Starkstromkonsument: I was inspired by SystemStatistics.php. It uses 'N/A' for all undefined values. I prefer returning a defined value for monitoring systems. It makes it easier to distinguish between a value not being available ('N/A') or some other error (empty values).

Starkstromkonsument:
@kesselb thank you for your review. I use this information to add triggers and graphs for memory usage, hit rates and number of cached variables / scripts. I don't like software components running as black boxes ;-) I need the actual values and the configured maximum values too to be able to configure triggers by percentages of free/usage.

lib/PhpStatistics.php
kesselb: I don't see how mixing types (string & array) makes it easier to distinguish.

lib/PhpStatistics.php
kesselb: I think it's fine. The method seems to return false if disabled.

kesselb:
> I use this information to add triggers and graphs for memory usage, hit rates and number of cached variables / scripts. I don't like software components running as black boxes ;-) I need the actual values and the configured maximum values too to be able to configure triggers by percentages of free/usage.

That's not a reason to add this information to the serverinfo api ;)

Starkstromkonsument:
I'm sorry, I don't get your point. What change to the PR do you suggest?

kesselb:
See my comments ;)

lib/PhpStatistics.php
```diff
@@ -64,4 +66,62 @@ protected function cleanVersion($version) {
...
+			return [];
+		}
+	}
+
+	/**
+	 * Get status information about the cache from the APCu extension
+	 *
+	 * @return array with an array of state information about the cache instance
+	 */
+	protected function getAPCuStatus(): array {
```
kesselb, line 98:
```
	/**
	 * Get status information about the cache from the APCu extension
	 *
	 * @return array with an array of state information about the cache instance
	 */
	protected function getAPCuStatus(): array {
		// Test if the APCu module is installed
		if (!extension_loaded('apcu')) {
			// module not loaded, returning back empty array to prevent any errors on JS side.
			return [
				'cache' => false,
				'sma' => false
			];
		}

		return [
			'cache' => apcu_cache_info(true),
			'sma' => apcu_sma_info(true)
		];
	}
```

Looks a bit easier to me. What do you think?

lib/PhpStatistics.php
```diff
@@ -64,4 +66,62 @@ protected function cleanVersion($version) {
...
+	protected function getOPcacheStatus(): array {
+		// Test if the OPcache module is installed
+		if (!extension_loaded('Zend OPcache')) {
+			// module not loaded, returning back empty array to prevent any errors on JS side.
+			return [];
+		}
+
+		// get status information about the cache
+		$status = opcache_get_status(false);
+
```
kesselb, line 84:
```
		// Test if the OPcache module is installed
		if (!extension_loaded('Zend OPcache')) {
			// module not loaded, returning back empty array to prevent any errors on JS side.
			return [];
		}

		// get status information about the cache
		$status = opcache_get_status(false);

		if ($status === false) {
			$status = [];
		}

		return $status;
```

I think it's much easier to read like this.

lib/PhpStatistics.php
Starkstromkonsument: Yes it does. But doesn't this result in mixing return types for parts of the array again?

kesselb: > Yes it does. But doesn't this result in mixing return types for parts of the array again?

That's okay as long as `getAPCuStatus` returns an array.

Starkstromkonsument: I would prefer to return 'N/A' instead of false in those cases then, as I proposed in the first place. What do you think?

kesselb: Fine by me.

PR #213, opened 2020-06-18
Move to new bootstrap mechanism

kesselb:
Much easier to register all those event listeners used by serverinfo now :rofl:


kesselb:
Seems to work ;)

ChristophWurst:
Wait … why was that even there? Did it do *anything at all*?

PR #215, opened 2020-06-19
Fix memory, Add Mbps/Gbps to Speed, Add Number of Cores and fix block-size on Storage/Swap Size (FreeBSD)

constrict:
#210 Additional changes to show correct values

### DIFF
--- a/lib/OperatingSystems/FreeBSD.php
+++ b/lib/OperatingSystems/FreeBSD.php
@@ -46,7 +46,7 @@ class FreeBSD {
 		$data = ['MemTotal' => -1, 'MemFree' => -1, 'MemAvailable' => -1, 'SwapTotal' => -1, 'SwapFree' => -1];
 
 		try {
-			$swapinfo = $this->executeCommand('/usr/sbin/swapinfo');
+			$swapinfo = $this->executeCommand('/usr/sbin/swapinfo -k');
 			$line = preg_split("/[\s]+/", $swapinfo);
 
 			if (count($line) > 3) {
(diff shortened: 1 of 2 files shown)


kesselb:
https://github.com/nextcloud/serverinfo/blob/481df199df1c93342c1f9f922a9b3badd9dd3fa3/lib/SystemStatistics.php#L115-L126

`hw.physmem` was also used with the old code. It's not correct?

constrict:
> `hw.physmem` was also used with the old code. It's not correct?

I suppose it depends on how you want to display the memory in two situations. The hw.realmem reports the reported real values of the memory totaled (in my case 64.00 GB). The hw.physmem displays the actual available memory that can be used by the system. So these are two different use case. I don't think either can be considered wrong in each situation.

The code in the FreeBSD.php is used to set the memory value in the image below. It does not appear to change any other value.

image

The old code is what sets the memory in the live usage display. FreeBSD.php has no influence on these values. SystemStatistics.php controls this image.
https://github.com/nextcloud/serverinfo/blob/481df199df1c93342c1f9f922a9b3badd9dd3fa3/lib/SystemStatistics.php#L115-L126

image

VVD:
Is any possibility to merge it to stable19 branch or even stable18?

kesselb:
> Is any possibility to merge it to stable19 branch or even stable18?

N as this a new feature.

VVD:
> N as this a new feature.

For me it's bug fix - system information was incorrect before this patch.

kesselb:
> system information was incorrect before this patch.

To be precise: System information was incorrect before this patch for the unsupported operating system FreeBSD. If that is such a problem for you just disable the serverinfo app ;)

VVD:
:-P

Make FreeBSD supported operating system. :+1:

PR #217, opened 2020-06-20
Use OS::getMemory for information about memory and swap.

kesselb:
To do:

- [x] Use memory object
- [ ] Update: [script.js](https://github.com/nextcloud/serverinfo/blob/f71748d5d8f530068a5ea3f1bd9a03eea3c2405b/js/script.js#L159-L167)

### DIFF
--- a/lib/SystemStatistics.php
+++ b/lib/SystemStatistics.php
@@ -66,10 +66,10 @@ class SystemStatistics {
 			'debug' => $this->config->getSystemValue('debug', false) ? 'yes' : 'no',
 			'freespace' => $this->getFreeSpace(),
 			'cpuload' => $processorUsage['loadavg'],
-			'mem_total' => $memoryUsage['MemTotal'],
-			'mem_free' => $memoryUsage['MemFree'],
-			'swap_total' => $memoryUsage['SwapTotal'],
-			'swap_free' => $memoryUsage['SwapFree'],
+			'mem_total' => $memoryUsage->getMemTotal() * 1024,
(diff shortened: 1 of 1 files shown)


come-nc:
@kesselb Is a lot of work left? This looks like a good change which might fix issues.

kesselb:
@come-nc I don't remember :see_no_evil: 

>  Use memory object

That's about this one: https://github.com/nextcloud/serverinfo/blob/master/lib/Resources/Memory.php

getMemory returns a Memory object now and the above calls needs to be updated. 

> Update: script.js

I think there is (or was) some normalization logic related to memory in script.js. Not sure what needs to be updated there.

kesselb:
@come-nc I don't remember :see_no_evil: 

>  Use memory object

That's about this one: https://github.com/nextcloud/serverinfo/blob/master/lib/Resources/Memory.php

getMemory returns a Memory object now and the above calls needs to be updated. 

> Update: script.js

I think there is (or was) some normalization logic related to memory in script.js. Not sure what needs to be updated there.

come-nc:
Note: previous code had a special case handling for using MemFree instead of MemAvailable if MemAvailable is not there because of an older linux kernel. MemAvailable was added in linux 3.14, I guess we do not need support for older kernel than this anymore?

CarlSchwan (approved):

kesselb:
>  I guess we do not need support for older kernel than this anymore?

Should be fine yes. Linux 3.14 was released in 2014.

PR #222, opened 2020-08-10
Add memory object and interface

kesselb:


lib/OperatingSystems/IOperatingSystem.php
```diff
@@ -0,0 +1,51 @@
...
+	 * @return Memory
+	 */
+	public function getMemory(): Memory;
+
+	/**
+	 * Get name of the processor
+	 *
+	 * @return string
+	 */
+	public function getCPUName(): string;
```
ChristophWurst, line 43:
```suggestion
	public function getCpuName(): string;
```

debatable, but I think that reads better

lib/Resources/Memory.php
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\ServerInfo\Resources;
+
+class Memory {
+	public $memTotal = -1;
```
ChristophWurst, line 29: how about setters and getters and `null` for undefined values?

lib/OperatingSystems/IOperatingSystem.php
kesselb: Fine by me.

lib/Resources/Memory.php
kesselb: done

PR #231, opened 2020-09-05
Normalize memory and disk values to MB

kesselb:
Close #230 

`Argument 1 passed to OCA\\ServerInfo\\Resources\\Disk::setUsed() must be of the type int, float given`

Disk and Memory objects are strictly typed. That's a problem if you are using more than 2147483647 bytes on a 32 bit system. To workaround this issue I changed the units to MB and normalize the values before.  As hardening the last call is always `(int)`.

PR #235, opened 2020-09-17
Update calculation for used and available disk space in gigabyte.

### LINKED ISSUE #234: System view in settings still partly broken
The charts that show the utilisation of the mount points is missing. There should be a chart in every box that shows how full the mount point is. 
Screen Shot 2020-09-17 at 11 29 07
Tested in Chrome and Safari. 
Using Nextcloud 20.0.0 beta 4

kesselb:
Fix #234 

After #231 the values returned by getUsed and getAvailable are already in megabyte.


MorrisJobke (approved):
🐘

PR #242, opened 2020-09-29
[TimeServer]: Updated the function to mainly use PHP only

HigH-HawK:
Added a new function to check for processes (get returnCode and PID) - This is being used in the getTimeServers() function, to query the correct timeserver process and to use the corresponding config, to fetch the timeservers in use.


HigH-HawK:
@kesselb Please could you review the changes and submit them for merge, if you are ok with them? 

I can't seem to signoff the commit, as there seem to be 5 unsigned commits. Not sure what went wrong but my Git Desktop App did not play ball for some reason and kept changing and recommiting the file.

kesselb:
Thanks @HigH-HawK :+1: 

I had in mind to remove the timeserver's because this information is not important (at least I think it's not important) ;) Also there are some ways to keep your time up to date. We need a lot of code to fetch the right information.

lib/OperatingSystems/DefaultOs.php
```diff
@@ -126,12 +126,71 @@ public function getUptime(): int {
 		return $uptimeInSeconds;
 	}
 
+	/**
+	 * @return string
+	 */
+	public function procExists($procName) {
+		exec("pgrep {$procName}", $output, $returnCode);
```
kesselb, line 133: `$this->executeCommand` should be used.

HigH-HawK:
Hey @kesselb 

I do agree and thought of that as well, while changing the function. There are many ways and programs to keep your time up to date and the code would need to grow a lot to get all of them under one roof. I wouldn't mind if the timeserver would disappear completely.

If you agree, we can close this pull request 👍

MorrisJobke:
Fixed with #243

PR #243, opened 2020-10-03
Remove timeserver

kesselb:
Close #188 #242 

Configured timeservers is a less important information. The current implementation show the timeservers only for some setups. For example chrony is not supported. On my Ubuntu box there are not timeservers shown because the defaults are used.


kesselb:
cc @HigH-HawK @j-ed @tflidd as you were involved in the other pull requests.

MorrisJobke (approved):
🐘

PR #246, opened 2020-10-07
Fix num_shares_link_no_password for NC >= 15

orandev:
It says in https://github.com/nextcloud/server/blob/master/apps/files_sharing/lib/Controller/ShareAPIController.php:
"share_with" and "share_with_displayname" for passwords of link shares was deprecated in Nextcloud 15, use "password" instead.


kesselb (approved):
Awesome :sunglasses:

Issue #250, opened 2020-10-09
Settings: System not working


vanastasis:
Hi,

Settings: system not displaying CPU, RAM etc...

error:

Internal Server Error

The server was unable to complete your request.

If this happens again, please send the technical details below to the server administrator.

More details can be found in the server log.
Technical details

    Remote Address: 10.1.1.1
    Request ID: oGZYlBRjDVd8swftuniE

Running; 
OS : TrueNas 12 / FreeBSD 12.2
Jail base : 12.2 FreeBSD
Jail nextcloud : 20
php : 4.7.10
web server : nginx 1.18
database : mysql 5.7.31
Install : fresh via plugins

index.php/settings/integrity/failed  = no errors

Enabled Apps:
  - accessibility: 1.6.0
  - activity: 2.13.1
  - cloud_federation_api: 1.3.0
  - dav: 1.16.0
  - federatedfilesharing: 1.10.1
  - files: 1.15.0
  - files_rightclick: 0.17.0
  - files_sharing: 1.12.0
  - files_trashbin: 1.10.1
  - files_versions: 1.13.0
  - logreader: 2.5.0
  - lookup_server_connector: 1.8.0
  - nextcloud_announcements: 1.9.0
  - notifications: 2.8.0
  - oauth2: 1.8.0
  - password_policy: 1.10.1
  - photos: 1.2.0
  - privacy: 1.4.0
  - provisioning_api: 1.10.0
  - serverinfo: 1.10.0
  - settings: 1.2.0
  - survey_client: 1.8.0
  - theming: 1.11.0
  - twofactor_backupcodes: 1.9.0
  - updatenotification: 1.10.0
  - viewer: 1.4.0
  - workflowengine: 2.2.0

Tested on 
- Windows 10, Firefox 82, IE 11, Edge
- Linux, Debian 12, Firefox 82

kesselb:
Mind to check the nextcloud logfile for error messages? Otherwise there is not much we can do. Could be https://github.com/nextcloud/serverinfo/pull/240 which is fixed with Nextcloud 20.0.1 (which is not released yet).

vanastasis:
Hi,

Thank you for the email

when i click on system, it creates 3 logs, see below.

[...]

PR #251, opened 2020-10-11
Fix FreeBSD Interface Exception

constrict:
#250 

Null is returned by a pregmatch not an empty string.

### DIFF
--- a/lib/OperatingSystems/FreeBSD.php
+++ b/lib/OperatingSystems/FreeBSD.php
@@ -184,8 +184,13 @@ class FreeBSD implements IOperatingSystem {
 				preg_match("/(?<=\<).*(?=-)/m", $intface, $duplex);
 
 				$iface['mac'] = implode(' ', $mac[0]);
-				$iface['status'] = $status[0];
 				$iface['speed']  = $speed[0];
+
+				if ($status[0] !== null) {
+					$iface['status'] = $status[0];
+				} else {
+					$iface['status'] = 'active';
+				}
 
 				if ($iface['speed'] !== null) {
 					if (strpos($iface['speed'], 'G')) {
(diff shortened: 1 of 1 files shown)


kesselb:
Are you able to reproduce this on your box? I would like to add unit tests at some point for the network logic (like we already have for memory and disks) so every breaking input is nice to have ;)

I'm tried one of the reports from the issue but the input seems to work: https://3v4l.org/lPiVo

constrict:
I was able to reproduce the issue by adding a bridge interface.

Checking for the null does work. I also added a default for a null status.

image

```
bridge0: flags=8802<BROADCAST,SIMPLEX,MULTICAST> metric 0 mtu 1500
        ether 00:00:00:00:00:00
        id 00:00:00:00:00:00 priority 32768 hellotime 2 fwddelay 15
        maxage 20 holdcnt 6 proto rstp maxaddr 2000 timeout 1200
        root id 00:00:00:00:00:00 priority 0 ifcost 0 port 0
        groups: bridge
        nd6 options=9<PERFORMNUD,IFDISABLED>

```

lib/OperatingSystems/FreeBSD.php
```diff
@@ -184,10 +184,15 @@ public function getNetworkInterfaces() {
 				preg_match("/(?<=\<).*(?=-)/m", $intface, $duplex);
 
 				$iface['mac'] = implode(' ', $mac[0]);
-				$iface['status'] = $status[0];
 				$iface['speed']  = $speed[0];
 
-				if ($iface['speed'] !== '') {
+				if ($status[0] !== null) {
```
kesselb, line 189:
```suggestion
				if (isset($status[0])) {
```

kesselb (approved):

lib/OperatingSystems/FreeBSD.php
constrict: Changed all three statements to isset.

PR #256, opened 2020-11-19
Properly fetch oracle database information

juliushaertl:
I'm not sure if the previous one ever worked, however querying `SELECT version FROM v$instance` doesn't on my test setup.

The additional catch block makes sure that we at least show something in case the query fails somewhere and not end up with an exception being thrown.

image

### DIFF
--- a/lib/DatabaseStatistics.php
+++ b/lib/DatabaseStatistics.php
@@ -22,6 +22,7 @@
 
 namespace OCA\ServerInfo;
 
+use Doctrine\DBAL\DBALException;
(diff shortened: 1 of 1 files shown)


kesselb (approved):
LGTM :+1:

ChristophWurst (approved):
I remember that we've had an instance with Oracle where this didn't work either and it was some kind of permission problem. Can't tell what's the better approach. :+1:?

kesselb:
Please don't forget to fix the code style ;)

PR #261, opened 2020-12-11
Fix MySQL database size calculation

nickvergessen:
Inspired by https://github.com/phpbb/phpbb/commit/cd235dfd428afe001fc0e5ae191136566ec3fbe0 from @3D-I

### DIFF
--- a/lib/DatabaseStatistics.php
+++ b/lib/DatabaseStatistics.php
@@ -95,12 +95,13 @@ class DatabaseStatistics {
 		// This code is heavily influenced by a similar routine in phpMyAdmin 2.2.0
 		switch ($this->config->getSystemValue('dbtype')) {
 			case 'mysql':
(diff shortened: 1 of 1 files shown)


lib/DatabaseStatistics.php
```diff
@@ -95,12 +95,13 @@ protected function databaseSize() {
 		// This code is heavily influenced by a similar routine in phpMyAdmin 2.2.0
 		switch ($this->config->getSystemValue('dbtype')) {
 			case 'mysql':
+				$mysqlEngine = ['MyISAM', 'InnoDB', 'Aria'];
```
juliushaertl, line 98: What about `MRG_MyISAM`, is that not relevant anymore?

juliushaertl: Ah sorry, misread the old check

juliushaertl (approved):

PR #267, opened 2021-01-11
Added 7 and 30 day values to "Active Users" stats graph

akhil1508:
- Added 7 and 30 day values to the "Active users" stats graph (returned through "lib/SessionStatistics.php" and used in frontend in "js/script.js)
- Modified unit tests accordingly to test that new values are correctly returned
- Package "nextcloud/coding-standard" updated to "^0.4.0" from "^0.3.0"(PHPLint refuses to install correctly with 0.3.0 in the checks otherwise)


arnauvp:
Ping

J0WI:
The performance is already pretty bad on large instances. Most likely due share statistics but better safe than sorry.

arnauvp:
On our instance, the `serverinfo/api/v1/info?format=json` call takes up to 5.5 seconds. The DB query to fetch unique users in last 30 days takes ~400ms before it's cached.

nickvergessen:
Needs a major overhaul after https://github.com/nextcloud/serverinfo/pull/420

Want to have a look?

PR #269, opened 2021-01-19
Allow API endpoint access with a predefined token

### LINKED ISSUE #100: User with limited rights dedicated only for monitoring
Hi,
I would like suggest to add user with limited rights (or user group) wich will be dedicated only for monitring. I want use XML with my external service but I am little afraid to use administrator account for it.
Thank you

PVince81:
Quick and dirty approach for accessing the metrics endpoint without admin credentials, not that user friendly but works.

Fixes #100

Drawbacks:
- it uses its own token approach (inspired by how the music app does it for Ampache). Making it use the proper token facility requires many more changes to be able to generate tokens with more restrictions (might need core/server changes)
- only one token for now
- not configurable in the web UI, need occ

Issues:
- [ ] unsure about the NC-Token header, need to check if there are other places where we use different headers
- [ ] header vs URL param (which could leak more easily)

### DIFF
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -28,14 +28,27 @@ use OCA\ServerInfo\SessionStatistics;
 use OCA\ServerInfo\ShareStatistics;
 use OCA\ServerInfo\StorageStatistics;
 use OCA\ServerInfo\SystemStatistics;
+use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\OCSController;
+use OCP\IConfig;
+use OCP\IGroupManager;
 use OCP\IRequest;
+use OCP\IUserSession;
 
 class ApiController extends OCSController {
 
 	/** @var Os */
 	private $os;
+
+	/** @var IConfig */
+	private $config;
+
+	/** @var IGroupManager */
+	private $groupManager;
+
+	/** @var IUserSession */
+	private $userSession;
 
 	/** @var SystemStatistics */
 	private $systemStatistics;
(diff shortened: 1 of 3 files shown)


lib/Controller/ApiController.php
```diff
@@ -88,12 +110,45 @@ public function __construct($appName,
...
+			return false;
+		}
+
+		$user = $userSession->getUser();
+		if ($user === null) {
+			return false;
+		}
+
+		if (!$this->groupManager->isAdmin($user->getUID())) {
+			throw new NotAdminException($this->l10n->t('Logged in user must be an admin'));
```
PVince81, line 133: oops

PVince81:
rebased for CI.

if this minimal approach is acceptable and secure enough I can then move forward to write tests for this

thoughts ? @LukasReschke @nickvergessen @kesselb

lib/Controller/ApiController.php
```diff
@@ -88,12 +110,45 @@ public function __construct($appName,
 		$this->sessionStatistics  = $sessionStatistics;
 	}
 
+	private function checkAuthorized() {
+		$token = $this->request->getHeader('NC-Token');
+		if (!empty($token)) {
+			$storedToken = $this->config->getAppValue('serverinfo', 'token', null);
+			if ($storedToken === $token) {
```
LukasReschke, line 117: This should use https://www.php.net/manual/en/function.hash-equals.php

PVince81: thanks, done

PVince81:
I've added unit tests

MorrisJobke (approved):
I fixed the info text (which said `OC-Token` instead of `NC-Token`). But then it works fine 👍

Issue #278, opened 2021-02-23
Internal server error when clicking the tab


ikomhoog:
I am a bit overwhelmed by the info that I get spewed at me, I'm also not a big user of Nextcloud.
If I need to provide anything else, please explain where I can get it.
Apps also wouldn't update so my install might just be broken.

### Steps to reproduce
1. install app
2. try to look at settings -> system
3. internal server error

### Expected behaviour
get shown a page with system info?

### Actual behaviour
Tell us what happens instead

### Server configuration

**Operating system**: nux 5.10.1-Unraid #1 SMP Thu Dec 17 11:41:39 PST 2020 x86_64

**Web server:**  Apache/2.4.38 (Debian) (apache2handler)

**Database:** sqlite3 3.27.2

**PHP version:** 7.4.15
Modules loaded: Core, date, libxml, openssl, pcre, sqlite3, zlib, ctype, curl, dom, fileinfo, filter, ftp, hash, iconv, json, mbstring, SPL, PDO, session, posix, Reflection, standard, SimpleXML, pdo_sqlite, Phar, tokenizer, xml, xmlreader, xmlwriter, mysqlnd, apache2handler, apcu, bcmath, exif, gd, gmp, imagick, intl, ldap, memcached, pcntl, pdo_mysql, pdo_pgsql, redis, sodium, zip, Zend OPcache

**Nextcloud version:** (see Nextcloud admin page) 21.0.0 - 21.0.0.18

**Where did you install Nextcloud from:** "unknown" according to the server configuration detail, but its docker

**List of activated apps:** 

```
Enabled:
 - accessibility: 1.7.0
 - activity: 2.14.3
 - admin_audit: 1.11.0
 - cloud_federation_api: 1.4.0
 - dashboard: 7.1.0
 - dav: 1.17.1
 - deck: 1.3.0
 - federatedfilesharing: 1.11.0
 - files: 1.16.0
 - files_pdfviewer: 2.1.0
 - files_rightclick: 1.0.0
 - files_sharing: 1.13.1
 - files_trashbin: 1.11.0
 - files_versions: 1.14.0
 - files_videoplayer: 1.10.0
 - firstrunwizard: 2.10.0
 - logreader: 2.6.0
 - lookup_server_connector: 1.9.0
 - nextcloud_announcements: 1.10.0
 - notifications: 2.9.0
 - oauth2: 1.9.0
 - password_policy: 1.11.0
 - photos: 1.3.0
[...]

PR #283, opened 2021-02-27
fix Internal Server Error @ /settings/admin/serverinfo in 21.0.0

vszakats:
Fixes: https://github.com/nextcloud/serverinfo/issues/278


rullzer (approved):

MorrisJobke:
I just looked into this before lunch and fixed it after just to realize that it is now fixed 😢

MorrisJobke:
> I just looked into this before lunch and fixed it after just to realize that it is now fixed 😢

But good that it's fixed 👍

PR #288, opened 2021-03-27
Use strict typing

J0WI:


lib/DatabaseStatistics.php
```diff
@@ -88,9 +91,9 @@ protected function databaseVersion() {
 	 * @copyright (c) phpBB Limited <https://www.phpbb.com>
 	 * @license GNU General Public License, version 2 (GPL-2.0)
 	 *
-	 * @return int|string
+	 * @return string
 	 */
-	protected function databaseSize() {
+	protected function databaseSize(): string {
```
blizzz, line 96: The method might return `$database_size`  as an int  currently. Probably cast it in the return statement.

blizzz (changes requested):
one method returns mixed yet

lib/DatabaseStatistics.php
J0WI: fixed

lib/Os.php
```diff
@@ -44,45 +47,59 @@ public function __construct() {
...
+	/**
+	 * @return Memory
+	 */
 	public function getMemory(): Memory {
 		return $this->backend->getMemory();
 	}
 
+	/**
+	 * @return string
+	 */
```
blizzz, line 80: when we have return type hints, we don't need phpdoc just for this. not a blocker, for next time.

blizzz:
still needs a rebase, there's a conflict with  lib/StorageStatistics.php

blizzz (approved):

PR #292, opened 2021-04-02
Update settings-admin.php

HouraisanNEET:
Change the position of ``<em id="numUsersStorage">``.
## Before
Screenshot from 2021-04-02 20-17-31
## After
Screenshot from 2021-04-02 20-18-25


come-nc:
@jancborchardt Any opinion?
I’m not sure I see the point.

Dennis1993:
In my test installation the change is already available:
image

If this is not correct please open this pull request or write me.

PR #293, opened 2021-04-06
Add L10N to formatUptime()

HouraisanNEET:
Add L10N to ``formatUptime()``

Signed-off-by: HouraisanNEET


kesselb:
Thanks :+1: 

I think it's fine to not translate the uptime.

PR #297, opened 2021-04-22
move counting storage statistics to the background

blizzz:
Previously we had a `COUNT(*)` against the filecache table. At least on MySQL's InnoDB this caused a full table scan and could result on MySQL quitting for a too long running query. Since it was done live it also meant long loading times for the web interface as well as API call.

This change moves the counting into the background and uses a cached value only. By default it is being updated every three hours, but can be configured via app config (example added in README). Further, it can be triggered manually via occ command (example also in README).

### DIFF
--- a/lib/StorageStatistics.php
+++ b/lib/StorageStatistics.php
@@ -22,27 +22,31 @@
 
 namespace OCA\ServerInfo;
 
+use OCP\IConfig;
 use OCP\IDBConnection;
 
 class StorageStatistics {
 
 	/** @var  IDBConnection */
 	private $connection;
+	/** @var IConfig */
+	private $config;
 
 	/**
 	 * SystemStatistics constructor.
 	 *
 	 * @param IDBConnection $connection
 	 */
-	public function __construct(IDBConnection $connection) {
+	public function __construct(IDBConnection $connection, IConfig $config) {
 		$this->connection = $connection;
+		$this->config = $config;
 	}
 
 	public function getStorageStatistics() {
 		return [
 			'num_users' => $this->countUserEntries(),
-			'num_files' => $this->countEntries('filecache'),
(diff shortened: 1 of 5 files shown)


lib/StorageStatistics.php
```diff
@@ -65,18 +69,34 @@ protected function countUserEntries() {
...
+
+		$storageQuery = $this->connection->getQueryBuilder();
+		$storageQuery->selectAlias('numeric_id', 'id')
+			->from('storages');
+		$storageResult = $storageQuery->execute();
+		while ($storageRow = $storageResult->fetch()) {
+			$storageCount++;
+			$fileQuery->setParameter('storageId', $storageRow['id']);
+			$fileResult = $fileQuery->execute();
+			$fileCount += (int)$fileResult->fetchOne();
```
kesselb, line 93: fetchOne is not available for 20 :(

nickvergessen: so adjust on the backport to fetchColumn()

blizzz: Yup, this here goes against 22. Backports get their care when due.

kesselb (approved):

lib/StorageStatistics.php
```diff
@@ -65,18 +69,34 @@ protected function countUserEntries() {
...
+	protected function getCountOf(string $table): int {
+		return (int)$this->config->getAppValue('serverinfo', 'cached_count_' . $table, '0');
+	}
+
+	public function updateStorageCounts(): void {
+		$storageCount = 0;
+		$fileCount = 0;
+
+		$fileQuery = $this->connection->getQueryBuilder();
+		$fileQuery->select($fileQuery->createFunction('COUNT(*)'))
```
nickvergessen, line 81:
```suggestion
		$fileQuery->select($fileQuery->func()->count())
```

Issue #291, opened 2021-03-30
Internal server error when ifconfig is unavailable (FreeBSD)


Tsuroerusu:
I run a multi-tenant Nextcloud server in production with very tight security restrictions. One of those restrictions is that I use the chroot feature of PHP-FPM to prevent the Nextcloud instances from looking around in the system environment.

Inside my chroot environment is a very very minimal folder structure with /dev and /etc only containing what is necessary for OpenSSL to function, and no binaries or libraries at all (PHP-FPM loads everything it needs before chrooting). This has worked brilliantly for several years at this point with the usual feature here and there being unavailable, but nothing actually failing or not working. In fact, I have not previously had any apps, including serverinfo, "hard fail" because of this.

I use FreeBSD as my operating system, however I could imagine something similar happening on Linux under the right circumstances. For example, if SELinux was preventing serverinfo from getting information about the network configuration.

### Steps to reproduce
1. Install Nextcloud on FreeBSD.
2. Delete or rename the "ifconfig" binary 
3. Open the serverinfo app.

### Actual behaviour
When I go to the serverinfo app's page in the admin panel, I get an internal server error and when I then go back to the log, I see that it is failing because it cannot invoke the **ifconfig** command which is, deliberately, unavailable inside my chroot as I see no need for my instances looking at the network configuration. In fact, it is only on a single shared instance, where I am the administrator, that I even have the serverinfo app enabled.

### Expected behaviour
In my opinion, this is rather backward. If the **ifconfig** binary is unavailable or gives the wrong output there should not be a hard error but rather, the features that relies on the output should simply be unavailable. In this case, I briefly tried running the instance with the chroot disabled and then serverinfo worked fine an showed a list of the available NICs in the system. Instead of giving an internal server error that stops everything from working (A hard error), the section where the list of NICs is shown should just say "Unavailable" and give a softer error in the log and thus not preventing the serverinfo app's other functions from working. At the very least it should be possible to disable the features relying on **ifconfig** in the case of FreeBSD or whatever network command is invoked on Linux which could be unavailable or restricted for all sorts of other reasons like SELinux policies. In my case, I also use FreeBSD jails which does not even use **ifconfig** inside the jail, making the binary even more unnecessary, as the network configuration is done by the host system before any services even start inside the jail.

### Server configuration

**Operating system:** FreeBSD 12.2

**Web server:** Apache 2.4.46

**Database:** MariaDB 10.4.17

**PHP version:** 7.3.27

[...]

PR #300, opened 2021-04-29
Catch RuntimeException

HouraisanNEET:
#291
Signed-off-by: HouraisanNEET <HouraisanNEET@users.noreply.github.com>

### DIFF
--- a/lib/OperatingSystems/FreeBSD.php
+++ b/lib/OperatingSystems/FreeBSD.php
@@ -17,7 +17,7 @@ declare(strict_types=1);
  * GNU Affero General Public License for more details.
  *
  * You should have received a copy of the GNU Affero General Public License, version 3,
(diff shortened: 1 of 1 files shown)


kesselb:
Hi :wave: 

Thanks for your patch. It's unnecessary to wrap the whole block in a try-catch statement. Please change it to

```
try {
    $ifconfig = $this->executeCommand('/sbin/ifconfig -a');
} catch (\RuntimeException $e) {
    return $result;
}
```

and inside the loop for the other executeCommand

```
try {
    $intface  = $this->executeCommand('/sbin/ifconfig ' . $iface['interface']);
} catch (\RuntimeException $e) {
    continue;
}
```

HouraisanNEET:
Okay, Fixed.

kesselb (approved):
Thank you :+1:

PR #319, opened 2021-06-13
Fix Chart.js Syntax

HouraisanNEET:
## Before
Before

## After
After


blizzz:
Is `yAxes` even still correct or should not be it just `y`?

HouraisanNEET:
Both ``yAxes`` and ``y`` are valid. And I keep ``yAxes`` for now.

blizzz (approved):
Thank you :)

PR #329, opened 2021-09-25
List php extensions

R0Wi:
In many situations it's good to know which php extensions are enabled in the current setup. Usually one would look this up in `phpinfo` page but this one is not included in NC, so i thought it would be nice to add this info to the serverinfo page?

### DIFF
--- a/lib/PhpStatistics.php
+++ b/lib/PhpStatistics.php
@@ -120,7 +120,7 @@ class PhpStatistics {
 	 * @return string as csv list of loaded extensions
 	 */
 	protected function getLoadedPhpExtensions(): string {
-		$extensions = get_loaded_extensions();
+		$extensions = (function_exists('get_loaded_extensions') ? get_loaded_extensions() : ['Unable to list extensions']);
 		return implode(', ', $extensions);
 	}
 }
--- a/lib/PhpStatistics.php
+++ b/lib/PhpStatistics.php
@@ -52,6 +52,7 @@ class PhpStatistics {
 			'upload_max_filesize' => $this->phpIni->getBytes('upload_max_filesize'),
 			'opcache' => $this->getOPcacheStatus(),
 			'apcu' => $this->getAPCuStatus(),
+			'extensions' => $this->getLoadedPhpExtensions(),
 		];
 	}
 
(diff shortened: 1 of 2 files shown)


R0Wi:
Sorry did some mistake, `rebased` on `master` :+1:

lib/PhpStatistics.php
```diff
@@ -112,4 +113,14 @@ protected function getAPCuStatus(): array {
...
 		];
 	}
+
+	/**
+	 * Get all loaded php extensions
+	 *
+	 * @return string as csv list of loaded extensions
+	 */
+	protected function getLoadedPhpExtensions(): string {
+		$extensions = get_loaded_extensions();
```
CarlSchwan, line 123: can you add a check if the function exists or use `@get_loaded_extensions();` ? This sounds like a function that can be disabled on shared hosts.

R0Wi: Good point 👍 According to https://www.php.net/manual/en/ini.core.php, any internal PHP function can be disabled via `disable_functions`. Having that in mind would it also make sense to add a `@` to other functions inside this class like `opcache_get_status`, `apcu_cache_info` or `apcu_sma_info`?

come-nc: @ operator will not work.
Please add change like this:
```suggestion
		$extensions = (function_exists('get_loaded_extensions') ? get_loaded_extensions() : ['Unable to list extensions']);
```

R0Wi (commented):
LGTM 🚀

come-nc (approved):

lib/PhpStatistics.php
```diff
@@ -112,4 +113,14 @@ protected function getAPCuStatus(): array {
...
 		];
 	}
+
+	/**
+	 * Get all loaded php extensions
+	 *
+	 * @return string as csv list of loaded extensions
+	 */
+	protected function getLoadedPhpExtensions(): string {
+		$extensions = (function_exists('get_loaded_extensions') ? get_loaded_extensions() : ['Unable to list extensions']);
```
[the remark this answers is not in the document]
R0Wi: Sounds cleaner to me, too. Shall i implement this task as separate PR? Or anyone else interested?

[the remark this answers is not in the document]
R0Wi: Done at https://github.com/nextcloud/serverinfo/pull/367

PR #379, opened 2022-05-21
Prevent PHP errors when iterating network interfaces

PhrozenByte:
glob() might return false, causing "foreach() argument must be of type array|object, bool given" PHP errors


kesselb (approved):
Thank you :+1:

come-nc (approved):

PR #415, opened 2023-01-20
chore(deps): drop PHP 7.4, add PHP 8.2

kesselb:
Drop PHP 7.4

- [x] Remove required check "php7.4-oci-master"
- [x] Remove required check "Lint / php7.4 (pull_request)"


composer.json
```diff
@@ -19,7 +19,7 @@
 			"composer/package-versions-deprecated": true
 		},
 		"platform": {
-			"php": "7.4"
+			"php": "8.0"
```
nickvergessen, line 22: If you run `composer update php` it will not update all dependencies but leave it to configured dependabot 🙈

nickvergessen (commented):
Don't want to go with the templated files from the .github repo? They'd have a summary we could require instead

composer.json
kesselb: Good one!

kesselb:
> Don't want to go with the templated files from the .github repo? They'd have a summary we could require instead

Works for me. Let me update the PR.

szaimen (commented):
LGTM and seems to work in my testing. 
- However the first opening of the web interface is pretty slow. But not sure if that is related. 
- Also, can we also add phpunit tests for mariadb and postgres?
- Please note that I didn't test the `ocs/v2.php/apps/serverinfo/api/v1/info` endpoint.

nickvergessen:
Bumped all templates
And then adjusted the required

kesselb:
> Bumped all templates And then adjusted the required

Thank you :+1:

nickvergessen (approved):

Issue #381, opened 2022-06-01
Add the ability to hide all irrelevant disks/partitions and network interfaces under settings/admin/serverinfo


Sommerwiesel:

Sommerwiesel:
Hi, thanks for the information.
I will do exactly that.

proxyconcept:
I'm also interested in this feature: we can have several mount points on the server that doesn't concern Nextcloud (and people who have access on this admin page are not sysadmins).

So I implemented an another optional filter: It can be activated in the system config. It's only a simple boolean, because the filtering is automatically based on the mount point: the disks is relevant if the mount path is parent of the PHP files or the data directory.

What do you think about that ?

PR #426, opened 2023-02-23
Relevant disks filtering (from nextcloud paths)

proxyconcept:
Another disks filter:
- optional: only if activated from the system config (new key `serverinfo.filter_disks`)
- simple to configure: just a boolean value (off by default)
- automatic relevance determination if the disks mount point is parent of the Nextcloud instance (html or data directory)

Reference #381

### DIFF
--- a/lib/OperatingSystems/DefaultOs.php
+++ b/lib/OperatingSystems/DefaultOs.php
@@ -209,6 +209,17 @@ class DefaultOs implements IOperatingSystem {
 			} elseif (in_array($matches['Mounted'][$i], ['/etc/hostname', '/etc/hosts'], false)) {
 				continue;
 			}
+			
+			// Filter system option: only relevant disks (mount point parent of html or data directory)
+			if (\OC::$server->getSystemConfig()->getValue('serverinfo.filter_disks', false)) {
+				$len = strlen($matches['Mounted'][$i]);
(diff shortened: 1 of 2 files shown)


kesselb:

Sorry, it took so long to review your pull request :see_no_evil: 

My idea for #381 is a bit different: A way to filter the disks via expressions (sda1 - exact match, sda* - match everything starting with sda). 

On my server, your idea would work as well. 
I cannot estimate whether this approach is useful or not. 

From technical point, we need to rework the feature a bit. We do some filtering in DefaultOS and FreeBSD, but that is to normalize the data. Your filter is something that should be done afterward to not duplicate the code.

https://github.com/nextcloud/serverinfo/blob/02cb07850239c5ca7271e6d44ec8b84263c70240/lib/Os.php#L78

Should be a good place. You can inject the IConfig via Dependency Injection to read the configuration value. In addition the configuration key should be self explaining. When you read "filter_disks" you know what it does but not how. Something like "filter_disks_by_datadirectory" maybe?

I know that everyone has a different idea how to configure / set up the disks. 
So this feature could be useful for some.

kesselb:
Thank you for the pull request :+1: 

I am closing it due to the lack of activity. If you have the time to continue, please make a new one.

PR #443, opened 2023-05-04
With FreeBSD jails, networkinfo->gateway can be empty

cyrilst:
When installed in a FreeBSD jail, this error appears in logs:
`Undefined array key "gateway" at /path/to/nextcloud/apps/serverinfo/templates/settings-admin.php#199`

The patch defines the array key with an empty string.


kesselb (approved):

kesselb:
Thanks :+1:

PR #447, opened 2023-05-13
Use the color-primary-element* variables

szaimen:
Explanation: the color-primary variables are not to be used in components because the introduce problems with high-contrast primary colors. Fix this by using the primary-element variables instead.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -73,7 +73,7 @@
 }
 
 .info-color-label--used::before {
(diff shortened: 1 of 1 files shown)


kesselb (approved):

szaimen:
Github is breaking. Merged with https://github.com/nextcloud/serverinfo/commit/60241cc04ebd547e4299ae74adef880ee98720a0

Issue #347, opened 2021-12-13
Error in logs


Valdnet:
There is an error in the logs:

```
Error	no app in context	Exception: Return value of OCA\ServerInfo\OperatingSystems\DefaultOs::getTime() must be of the type string, null returned in file '/home/xxx/public_html/xxx/apps/serverinfo/lib/OperatingSystems/DefaultOs.php' line 115

    /home/xxx/public_html/xxx/lib/private/AppFramework/App.php - line 157:

    OC\AppFramework\Http\Dispatcher->dispatch(OCA\ServerIn ... {}, "BasicData")

    /home/xxx/public_html/xxx/lib/private/Route/Router.php - line 302:

    OC\AppFramework\App::main("OCA\\Server ... r", "BasicData", OC\AppFramew ... {}, { _route: "o ... "})

    /home/xxx/public_html/xxx/ocs/v1.php - line 62:

    OC\Route\Router->match("/ocsapp/app ... a")

    /home/xxx/public_html/xxx/ocs/v2.php - line 23:

    require_once("/home/xxx/p ... p")

Spowodowany TypeError: Return value of OCA\ServerInfo\OperatingSystems\DefaultOs::getTime() must be of the type string, null returned

    /home/xxx/public_html/xxx/apps/serverinfo/lib/Os.php - line 89:

    OCA\ServerInfo\OperatingSystems\DefaultOs->getTime()

    /home/xxx/public_html/xxx/apps/serverinfo/lib/Controller/ApiController.php - line 179:

    OCA\ServerInfo\Os->getTime()

    /home/xxx/public_html/xxx/lib/private/AppFramework/Http/Dispatcher.php - line 217:

    OCA\ServerInfo\Controller\ApiController->BasicData()

    /home/xxx/public_html/xxx/lib/private/AppFramework/Http/Dispatcher.php - line 126:

    OC\AppFramework\Http\Dispatcher->executeController(OCA\ServerIn ... {}, "BasicData")

    /home/xxx/public_html/xxx/lib/private/AppFramework/App.php - line 157:

    OC\AppFramework\Http\Dispatcher->dispatch(OCA\ServerIn ... {}, "BasicData")

    /home/xxx/public_html/xxx/lib/private/Route/Router.php - line 302:

    OC\AppFramework\App::main("OCA\\Server ... r", "BasicData", OC\AppFramew ... {}, { _route: "o ... "})

    /home/xxx/public_html/xxx/ocs/v1.php - line 62:

[...]

Issue #169, opened 2020-01-06
shell_exec


MeiRos:
Is it possible to not use shell_exec because it's unsafe? There are people who have disabled it and that's causing errors. Of course that also means something isn't working.

_error shell_exec() has been disabled for security reasons_

### Steps to reproduce
1.Install NC18.0.0 RC1 with PHP 7.4.1
2.Open the NC and go to the settings/system information
3.See log

### Expected behaviour
No errors

### Actual behaviour
I got errors and for example I can't see network information

### Server configuration

**Operating system:** Centos 7.8

**Web server:** Nginx 1.17.7

**Database:** MariaDB 10.3.21
**PHP version:** 7.4.1

**Nextcloud version:** Nextcloud 18.0.0 RC1

**Updated from an older Nextcloud/ownCloud or fresh install:** fresh

**Where did you install Nextcloud from:** download.nextcloud.com

**Signing status:**
<details>
<summary>Signing status</summary>

```
No errors have been found.
```
</details>

**List of activated apps:**<summary>Only default which are tested</summary>

**Are you using external storage, if yes which one:** no

**Are you using encryption:** no

**Are you using an external user-backend, if yes which one:** no

#### Nextcloud log (data/nextcloud.log)
<details>
<summary>Nextcloud log</summary>

```
Insert your Nextcloud log here
{"reqId":"rgMe4fuvcWCRhoGuelTv","level":3,"time":"2020-01-04T21:14:33+00:00","remoteAddr":"x","user":"admin","app":"PHP","method":"GET","url":"/ocs/v2.php/apps/serverinfo/api/v1//basicdata?format=json","message":"shell_exec() has been disabled for security reasons at /home/nginx/home.net/public/apps/serverinfo/lib/OperatingSystems/DefaultOs.php#96","userAgent":"Mozilla/5.0 (X11; CrOS x86_64 12607.58.0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/79.0.3945.86 Safari/537.36","version":"18.0.0.8","id":"5e1101566eed7"}

[...]

PR #448, opened 2023-05-13
Get system date/time without needing shell_exec()

joshtrichards:
One less use of shell_exec(). Maintains same output format typically used by `date` so no changes needed elsewhere.

Relevant to #169 and possibly #347.

It's also about 200x faster per call according to benchmarking via PHP's hrtime() on my test system, but that's 100% irrelevant unless one has _a lot_ of admins checking their System info simultaneously. 😆

### DIFF
--- a/lib/OperatingSystems/DefaultOs.php
+++ b/lib/OperatingSystems/DefaultOs.php
@@ -120,7 +120,7 @@ class DefaultOs implements IOperatingSystem {
 	}
 
 	public function getTime(): string {
-		return (string)shell_exec('date');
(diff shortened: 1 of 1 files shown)


kesselb:
The linux command `date` is local aware. 
Your implementation is not local aware anymore. 

```
LC_ALL=en_US.utf8 date
LC_ALL=de_DE.utf8 date
```

joshtrichards:
Oops completely overlooked that. Have a couple ideas. I'll try to take a look later in week.

Issue #362, opened 2022-03-30
[Enhancement] System Info


YouveGotMeowxy:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**

1. It would be very helpful to add the results of `<?php phpinfo(); ?>` to the "System" page
2. It would also be useful to add the cache info to that page.

**Describe the solution you'd like**

1. Just want a quick and easy way to view `phpinfo()` w/out having to manually handle it with my own custom page that I always have to rename when not in use, and name when I need it. Since NC is already secured via credentials, having this info on the system page shouldn't be considered a risk and more than your personal pics and docs on NC.
2. I recently changed cache type from APC to Redis, and it would be very handy to see the type(s) of cache NC is recognizing, usage stats of the cache(s) and so on. Currently the System tab is fairly basic and marginally useful; this would help toward making it more useful.

**Describe alternatives you've considered**

- Manually handling a `phpinfo()` page; not easy with the architecture of NC. Also more of a time waste, having to constantly rename the file when wanting/obscuring it.
- Creating an addon for the sole purpose of adding the `php phpinfo()`. Not being a coder, and especially not being at all familiar w/NC's architecture, it would take me WAY too long to do something so simple that an NC dev could probably add literally in 30 seconds. Also, I'd have no idea whatsoever on how to add caching details.

**Additional context**

Pretty plz w/sugar on top? :-p

joshtrichards:
Some of the caching details you may be looking for are already exposed through the serverinfo external monitoring endpoint. (You'll find the link at the bottom of your serverinfo page). Admittedly that's made for use by tools, but the data is there/gathered already. It can be accessed in both XML or JSON. That data includes - of what you requested:

- opcache statistics 
- NC's memcache local/distributed/locking configuration (though since this is a configured parameter I'm not sure how useful this is either way to expose)

In the interest of keeping the the user interface as clutter-free as possible, it might be good to explore what are the most important (and most widely useful) bits to add within the human facing interface (if any).

Personally I could see:

- opcache.revalidate_freq <-- in order to assess for server tuning as recommended in the NC Admin Manual
[...]

Issue #107, opened 2017-11-04
Add phpinfo() output


oparoz:
We have minimal information about the PHP configuration.
It would be good to have a hidden panel which can be expanded to reveal the full configuration.

Tarry91:
I want to second this idea! It would be really useful, especially because the nextcloud integrety check fails if one adds his own phpinfo.php file.

KopfKrieg:
It would also be nice to have phpinfo() available in the official docker container. Otherwise I have to modify the container, which is something I'd like to avoid. Having the output of phpinfo() somewhere in the settings would solve this problem.

dreua:
Any news on this topic or solutions you found in the meantime?

kesselb:
Implemented by https://github.com/nextcloud/serverinfo/pull/469

PR #462, opened 2023-05-31
Display full path(s) to currently active php .ini files

joshtrichards:
Helps people figure out which PHP ini file(s) are truly being used by their active NC installation

* Displays the full path(s) to the currently loaded `php.ini` file (if any) along with any other PHP configuration files parsed after `php.ini` (i.e. from additional compiled in paths or provided by the environment at run time).
* Furthers #362 and #107
* Desirable for:
  - Server Tuning
  - [Troubleshooting Web server and PHP problems
](https://docs.nextcloud.com/server/latest/admin_manual/issues/general_troubleshooting.html#troubleshooting-web-server-and-php-problems)
  - Upgrades
  - Migrations
  - Hardening and Security guidance
  - etc.

NC ServerInfo - Active PHP ini Files PR - Screenshot 2023-05-31 123228

(yellow just for screenshot)

### DIFF
--- a/lib/PhpStatistics.php
+++ b/lib/PhpStatistics.php
@@ -50,6 +50,7 @@ class PhpStatistics {
 			//   implementing a getOPcacheConfig() wrapper for PHP's opcache_get_configuration()
 			//   like we do for PHP's opcache_get_status() already below
 			'opcache' => $this->getOPcacheStatus(),
+			'php_ini_files' => $this->getInUsePhpIniFiles(),
 			'apcu' => $this->getAPCuStatus(),
 			'extensions' => $this->getLoadedPhpExtensions(),
 		];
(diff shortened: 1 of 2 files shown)


kesselb:
I don't like it.

joshtrichards:
You don't have to like it, you just have to find it useful. :)

Any particular reason?

If it's the presentation, I could display only the (unique) dirnames and drop the basenames entirely. This cuts out ~90% of the clutter (because in ~99% of installations most ini files are in only two unique local folders). 

It would still convey ~90% of the critical info IMO.

The info lost would be knowledge of whether a specific newly dropped in config file had become active or not (useful if someone isn't sure whether they've sufficiently restarted their web server, but not critical). Or a file in a configured path was being ignored due to permissions, a typo'd file extension, etc.

kesselb:
> you just have to find it useful. :)

I don't find it useful either.

Data created by PhpStatistics is automatically part of the monitoring endpoint. I don't see why someone would monitor the loaded php ini files.

joshtrichards:
Okay, fair. The same seems to apply to the existing PHP extensions list IMO. This sort of brings up the dichotomy between what belongs in serverinfo versus - well - elsewhere. 

Helping the admin figure out which PHP extensions are active and where the active PHP installation's config files aren't functions I'd typically expect to monitor via a constantly polled endpoint provided by NC. I'd consider them more diagnostic related.

Perhaps the Support app's "Generate system report" might be potential spot for this stuff. It's a shipped app, but doesn't appear to have a (public) repository though. (I also suspect it's more tailored towards what GmbH support wants for enterprise customer support matters (though it's useful to others). I don't know how much input the community has on what's included in it...

kesselb:

Closing the pull request as discussed.

PR #469, opened 2023-06-20
feat: add phpinfo

kesselb:


MichaIng (approved):
Good idea. This makes it easier for admins to check PHP configs, and whether theirs are applied as expected. I am thinking about OPcache, but helpful for other things as well.

Approving only the idea here, as I did not find time to test it yet.

templates/settings-admin.php
```diff
@@ -365,6 +365,9 @@ function FormatMegabytes(int $byte): string {
 							<?php p($l->t('Extensions:')); ?>
 							<em id="phpExtensions"><?php p($_['php']['extensions'] !== null ? implode(', ', $_['php']['extensions']) : $l->t('Unable to list extensions')); ?></em>
 						</p>
+						<p>
+							<a target="_blank" href="<?= $_['phpinfoUrl'] ?>">phpinfo</a>
```
szaimen, line 369: can we make this a button somehow? Currently it is not good to discover imho

szaimen (approved):
Tested and seems to work, just one small question :)

templates/settings-admin.php
szaimen: Or maybe use the primary color for it?

README.md
```diff
@@ -127,3 +127,7 @@ php occ serverinfo:update-storage-statistics -v --output=json_pretty
     "num_storages_other": 33
 }
 ```
+
+Show phpinfo
+
+``php occ config:app:set --value=yes serverinfo phpinfo``
```
szaimen, line 133: why are you making the feature opt-in instead opt-out?

szaimen: Maybe make it opt-out instead?

PR #517, opened 2023-11-01
feat: mark background job for storage statistics as time insensitive

kesselb:


joshtrichards:
My understanding of `TIME_INSENSITIVE` is that it's a no-op unless explicitly enabled by the admin (by setting `maintenance_window_start`)

This change may initially confuse a few people that happen to be using `maintenance_window_start` and are already polling storage stats, but seems mostly fine in terms of backwards compatibility. 

It's a little weird that if someone configures `maintenance_window_start` that `job_interval_storage_stats` will seem to essentially be ignored if the interval is >4 hours, but I also don't have a good alternative. Adding a warning to the `occ` command isn't practical since `maintenance_window_start` could get enabled after the fact. It may just be that we add a note to the doc (README currently) about the interaction between the two settings...

joshtrichards (approved):
Fine by me as long as a note is added to the README about interaction between `job_interval_storage_stats` and `maintenance_window_start`. - e.g.

> With the default serverinfo config, if `maintenance_window_start` is enabled then the default update interval (3 hours) becomes once per 24 hour period. If you customize the interval while using `maintenance_window_start` the interval you choose *must* be <4 hours or storage updates may never run at all.

kesselb:
Thank you, that's indeed a good point. 

I guess the [other jobs marked as time insensitive]( https://github.com/search?q=org%3Anextcloud+TIME_INSENSITIVE&type=code) are indeed different. For some users, it might be important to have recent values for files and storage available via monitoring. For now, we should therefore keep the current approach.

PR #572, opened 2024-03-15
Disks filtering by mount points from a configurable parameter

proxyconcept:
Here is my latest version from #426 

Use a new system config variable `serverinfo_disk_filter_paths` that can contains a list of paths (actually a simple string with `:` as delimiter). If defined, only disks with a corresponding mount point are shown.

The paths list can contains special name (`DOCROOT`, `DATADIR`, `TMPDIR` and `LOGSDIR`) which are automatically converted (example: `"serverinfo_disk_filter_paths" => "DATADIR:/var/log"`).

It works perfectly for my needs, I usually configure my Nextcloud instances with : `occ config:system:set --value="DATADIR" -- serverinfo_disk_filter_paths`


kesselb (commented):

JuliaKirschenheuter:
Hi @proxyconcept!

Thank you for your pull request! Would you like to finish small things which @kesselb suggested? Do you have an idea how to handle it? Or should we take over?

kesselb:
Closing due to inactivity.

Issue #534, opened 2023-11-27
Internal server error: OCA\ServerInfo\OperatingSystems\shell_exec()


Valdnet:
### Expected behaviour
Launching the serverinfo app.

### Actual behaviour
An internal server error occurs while running server information.

### Logs
```
[index] Error: Exception: Call to undefined function OCA\ServerInfo\OperatingSystems\shell_exec() in file '/home/xxx/public_html/cloud/apps/serverinfo/lib/OperatingSystems/DefaultOs.php' line 262 at <<closure>>

0. /home/xxx/public_html/cloud/lib/private/AppFramework/App.php line 183
   OC\AppFramework\Http\Dispatcher->dispatch(["OCA\\Settings\ ... "], "index")
1. /home/xxx/public_html/cloud/lib/private/Route/Router.php line 315
   OC\AppFramework\App::main("OCA\\Settings\\ ... r", "index", ["OC\\AppFramewo ... "], ["serverinfo","s ... "])
2. /home/xxx/public_html/cloud/lib/base.php line 1068
   OC\Route\Router->match("/settings/admin/serverinfo")
3. /home/xxx/public_html/cloud/index.php line 36
   OC::handleRequest()

Caused by:

Error: Call to undefined function OCA\ServerInfo\OperatingSystems\shell_exec() at <<closure>>

 0. /home/xxx/public_html/cloud/apps/serverinfo/lib/OperatingSystems/DefaultOs.php line 197
    OCA\ServerInfo\OperatingSystems\DefaultOs->executeCommand("df -TPk")
 1. /home/xxx/public_html/cloud/apps/serverinfo/lib/Os.php line 78
    OCA\ServerInfo\OperatingSystems\DefaultOs->getDiskInfo()
 2. /home/xxx/public_html/cloud/apps/serverinfo/lib/Settings/AdminSettings.php line 79
    OCA\ServerInfo\Os->getDiskInfo()
 3. /home/xxx/public_html/cloud/apps/settings/lib/Controller/CommonSettingsTrait.php line 129
    OCA\ServerInfo\Settings\AdminSettings->getForm()
 4. /home/xxx/public_html/cloud/apps/settings/lib/Controller/AdminSettingsController.php line 83
    OCA\Settings\Controller\AdminSettingsController->formatSettings([[["OCA\\ServerI ... ]])
 5. /home/xxx/public_html/cloud/apps/settings/lib/Controller/CommonSettingsTrait.php line 149
[...]

PR #598, opened 2024-05-06
fix: handle disabled shell_exec

kesselb:
Fix https://github.com/nextcloud/serverinfo/issues/534

### DIFF
--- a/lib/OperatingSystems/Linux.php
+++ b/lib/OperatingSystems/Linux.php
@@ -121,7 +121,11 @@ class Linux implements IOperatingSystem {
 	}
 
 	public function getTime(): string {
-		return (string)shell_exec('date');
+		try {
+			return $this->executeCommand('date');
+		} catch (RuntimeException $e) {
+			return '';
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


lib/OperatingSystems/Linux.php
```diff
@@ -139,12 +143,17 @@ public function getUptime(): int {
...
-		$result['dns'] = $dns;
-		$gw = shell_exec('ip route | awk \'/default/ { print $3 }\'');
-		$result['gateway'] = $gw;
+		$result = [
+			'hostname' => \gethostname(),
+			'dns' => '',
+			'gateway' => '',
+		];
+
+		if (function_exists('shell_exec')) {
```
joshtrichards, line 152: I guess the reason we can't use `executeCommand()` here is because of the inherent use of `escapeshellcmd()`. Maybe we add an optional parameter to `executeCommand()  that specifies whether to escape or not (that defaults true). We could then use consistently (elsewhere too I think) w/o boilerplate.

come-nc (approved):

PR #641, opened 2024-07-09
Use a more sensible format for load average

PeterFeicht:
The umpteen digits of precision in the default format are quite useless, even `uptime` shows only two decimal places. This makes it look a lot more sensible.

### DIFF
--- a/js/script.js
+++ b/js/script.js
@@ -111,7 +111,7 @@
 			});
 		}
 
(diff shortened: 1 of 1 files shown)


joshtrichards (approved):
Makes sense.

The chart itself already defaults to 2 for precision: https://github.com/nextcloud/serverinfo/blob/303ddee0ab92475b78a2db700b6c8990062926e0/js/smoothie.js#L306

kesselb (approved):
Thank you :+1:

PR #691, opened 2024-10-04
feat(active users): add more active user statistics intervals and improve presentation

tcitworld:
Adds:
* last 7 days
* last 1 month
* last 3 months
* last 6 months
* last year

to the API endpoint.

This view will certainly get converted to Vue at some point (and with proper CSS variables), but it felt like a nice improvement (and to use the new metrics).

### Before
Screenshot 2024-10-04 at 16-03-15 Système - Paramètres d'administration - Mon petit nuage

#### Full width
Screenshot 2024-10-04 at 16-00-56 System - Administration settings - Nextcloud
#### Tablet
Screenshot 2024-10-04 at 16-02-50 System - Administration settings - Nextcloud
#### Large Mobile
Screenshot 2024-10-04 at 16-01-45 System - Administration settings - Nextcloud
#### Small Mobile
Screenshot 2024-10-04 at 16-02-07 System - Administration settings - Nextcloud

### DIFF
--- a/templates/settings-admin.php
+++ b/templates/settings-admin.php
@@ -227,27 +227,39 @@ $phpinfo = $_['phpinfo'];
 
 			<div class="col col-12">
 				<div class="row">
-					<div class="col col-4 col-l-6 col-m-12">
+					<div class="col">
 						<div class="infobox">
-							<div class="interface-wrapper">
-								<?php if ($_['storage']['num_users'] > 0) : ?>
-									<?php p($l->t('Total users:')); ?>
-									<span class="info"><?php p($_['storage']['num_users']); ?></span><br>
+							<div class="interface-wrapper active-users-wrapper">
+								<?php if ($_['activeUsers']['last1hour'] > 0) : ?>
+									<div class="active-users-box">
+										<?php p($l->t('Last hour')); ?><br>
+										<span class="info"><?php p($_['activeUsers']['last1hour']) ?></span><br>
+										<em><?php p($l->t('%s%% of all users', [round($_['activeUsers']['last1hour'] * 100 / $_['storage']['num_users'], 1)])) ?></em>
+									</div>
 								<?php endif; ?>
 
 								<?php if ($_['activeUsers']['last24hours'] > 0) : ?>
-									<?php p($l->t('24 hours:')); ?>
-									<span class="info"><?php p($_['activeUsers']['last24hours']) ?></span><br>
+									<div class="active-users-box">
+										<?php p($l->t('Last 24 Hours')); ?><br>
+										<span class="info"><?php p($_['activeUsers']['last24hours']) ?></span><br>
+										<em><?php p($l->t('%s%% of all users', [round($_['activeUsers']['last24hours'] * 100 / $_['storage']['num_users'], 1)])) ?></em>
+									</div>
 								<?php endif; ?>
 
-								<?php if ($_['activeUsers']['last1hour'] > 0) : ?>
-									<?php p($l->t('1 hour:')); ?>
-									<span class="info"><?php p($_['activeUsers']['last1hour']) ?></span><br>
+								<?php if ($_['activeUsers']['last7days'] > 0) : ?>
+									<div class="active-users-box">
+										<?php p($l->t('Last 7 Days')); ?><br>
+										<span class="info"><?php p($_['activeUsers']['last7days']) ?></span><br>
+										<em><?php p($l->t('%s%% of all users', [round($_['activeUsers']['last7days'] * 100 / $_['storage']['num_users'], 1)])) ?></em>
+									</div>
 								<?php endif; ?>
 
-								<?php if ($_['activeUsers']['last5minutes'] > 0) : ?>
-									<?php p($l->t('5 mins:')); ?>
-									<span class="info"><?php p($_['activeUsers']['last5minutes']) ?></span><br>
+								<?php if ($_['activeUsers']['last1month'] > 0) : ?>
+									<div class="active-users-box">
+										<?php p($l->t('Last 30 Days')); ?><br>
+										<span class="info"><?php p($_['activeUsers']['last1month']) ?></span><br>
+										<em><?php p($l->t('%s%% of all users', [round($_['activeUsers']['last1month'] * 100 / $_['storage']['num_users'], 1)])) ?></em>
+									</div>
 								<?php endif; ?>
 							</div>
 						</div>
--- a/css/style.css
+++ b/css/style.css
@@ -259,3 +259,35 @@
 	align-items: center;
 	height: 24px;
 }
+
+.active-users-wrapper {
+	flex: 1;
+	display: flex;
+	gap: 0 1rem;
+	flex-wrap: wrap;
+	align-items: center;
+	justify-content: center;
+}
+
+.active-users-box {
+	padding: 0.25rem 1rem;
+	flex: 1;
+	min-width: 150px;
+	border-left: 2px solid var(--color-border);
+}
+
+@media (width <= 1280px) {
+	.active-users-box {
+		padding: 1rem 0.5rem;
+		border: none;
+		text-align: center;
+	}
+}
+
+.active-users-box:first-child {
+	border: none;
+}
+
+.active-users-box .info {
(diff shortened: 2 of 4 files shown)


nickvergessen:
I'm not affiliated with the app, so will not have time to review it.

templates/settings-admin.php
```diff
@@ -227,27 +227,46 @@ function FormatMegabytes(int $byte): string {
...
+					<div class="col">
 						<div class="infobox">
-							<div class="interface-wrapper">
-								<?php if ($_['storage']['num_users'] > 0) : ?>
-									<?php p($l->t('Total users:')); ?>
-									<span class="info"><?php p($_['storage']['num_users']); ?></span><br>
+							<div class="interface-wrapper active-users-wrapper">
+								<?php if ($_['activeUsers']['last1hour'] > 0) : ?>
+									<div class="active-users-box">
+										<?php p($l->t('1 Hour Active Users')); ?><br>
```
joshtrichards, line 235:
```suggestion
										<?php p($l->t('Last Hour')); ?><br>
```

templates/settings-admin.php
```diff
@@ -227,27 +227,46 @@ function FormatMegabytes(int $byte): string {
...
+										<span class="info"><?php p($_['activeUsers']['last1hour']) ?></span><br>
+										<em><?php p($l->t('%s%% of all users', [round($_['activeUsers']['last1hour'] * 100 / $_['storage']['num_users'], 1)])) ?></em>
+									</div>
 								<?php endif; ?>
 
 								<?php if ($_['activeUsers']['last24hours'] > 0) : ?>
-									<?php p($l->t('24 hours:')); ?>
-									<span class="info"><?php p($_['activeUsers']['last24hours']) ?></span><br>
+									<div class="active-users-box">
+										<?php p($l->t('1 Day Active Users')); ?><br>
```
joshtrichards, line 243:
```suggestion
										<?php p($l->t('Last 24 Hours')); ?><br>
```

templates/settings-admin.php
```diff
@@ -227,27 +227,46 @@ function FormatMegabytes(int $byte): string {
...
+										<em><?php p($l->t('%s%% of all users', [round($_['activeUsers']['last24hours'] * 100 / $_['storage']['num_users'], 1)])) ?></em>
+									</div>
 								<?php endif; ?>
 
-								<?php if ($_['activeUsers']['last1hour'] > 0) : ?>
-									<?php p($l->t('1 hour:')); ?>
-									<span class="info"><?php p($_['activeUsers']['last1hour']) ?></span><br>
+								<?php if ($_['activeUsers']['last7days'] > 0) : ?>
+									<div class="active-users-box">
+										<?php p($l->t('7 Day Active Users')); ?><br>
```
joshtrichards, line 251:
```suggestion
										<?php p($l->t('Last 7 Days')); ?><br>
```

templates/settings-admin.php
```diff
@@ -227,27 +227,46 @@ function FormatMegabytes(int $byte): string {
...
+										<span class="info"><?php p($_['activeUsers']['last7days']) ?></span><br>
+										<em><?php p($l->t('%s%% of all users', [round($_['activeUsers']['last7days'] * 100 / $_['storage']['num_users'], 1)])) ?></em>
+									</div>
+								<?php endif; ?>
+
+								<?php if ($_['activeUsers']['last1month'] > 0) : ?>
+									<div class="active-users-box">
+										<?php p($l->t('1 Month Active Users')); ?><br>
```
joshtrichards, line 259:
```suggestion
										<?php p($l->t('Last 30 Days')); ?><br>
```

templates/settings-admin.php
```diff
@@ -227,27 +227,46 @@ function FormatMegabytes(int $byte): string {
...
+										<em><?php p($l->t('%s%% of all users', [round($_['activeUsers']['last1month'] * 100 / $_['storage']['num_users'], 1)])) ?></em>
+									</div>
 								<?php endif; ?>
 
-								<?php if ($_['activeUsers']['last5minutes'] > 0) : ?>
-									<?php p($l->t('5 mins:')); ?>
-									<span class="info"><?php p($_['activeUsers']['last5minutes']) ?></span><br>
+								<?php if ($_['storage']['num_users'] > 0) : ?>
+									<div class="active-users-box">
+										<?php p($l->t('Total users')); ?><br>
```
joshtrichards, line 267: I feel like somehow the *Total users* should be separated out somehow since it's a bit of a different stat versus all the *Active Users* ones. It's more "Users that have ever logged in and still have an account in the database*. And it's a timeless stat so "Total" doesn't seem like the correct wording here.

Maybe just relabeling it *Users That Have Ever Logged In* but there must be a shorter phrase than I'm coming up with at the moment. :thinking:

joshtrichards:
A nice improvement, btw.

joshtrichards:
Just noticed this drops the 5 minute stat from the UI. Was that intentional?

templates/settings-admin.php
tcitworld: I feel like keeping the `Active users` suffix would make it more comprehensible. Maybe ask @nextcloud/designers ?

tcitworld:
> Just noticed this drops the 5 minute stat from the UI. Was that intentional?

I made the choice to not have too many statistics displayed in the UI, and I think it made more sense between 5 minutes and one hour to keep the second one.

kesselb:
image

I think the "last x" version is nicer. 
The headline is active users and therefore the context should be clear ;)

templates/settings-admin.php
kesselb: True. 

Total Users here are all users who logged in at least once. 

Okay for me to remove total users.

kesselb:
Thank you, nice enhancement :+1:

templates/settings-admin.php
tcitworld: 3 propositions : 
- simply rename to « All users »
- only show the number of users in the percentage subline (something like « xx% of all yyyy users », or « xx% of all users (yyyy) ». But this make this statistic not very visible.
- Move this stat to another block, with the possibility to differentiate between users who never logged once

tcitworld:
> I think the "last x" version is nicer.

Changed to this.

templates/settings-admin.php
kesselb: For now, I suggest we take this out and add an issue for thorough planning. I think we need a background job to fetch and cache the correct user count because looping over all backends is expensive. That only works for database/ldap backend, for saml/oidc we only know how many users were logged in once.

tcitworld: > I suggest we take this out

You mean remove the metric from this page altogether or simply show it elsewhere?

tcitworld: Done

jancborchardt (approved):
Awesome, looks really good!

Separate issue probably, but the icon of the heading is misaligned?

maximelehericy:
> > Just noticed this drops the 5 minute stat from the UI. Was that intentional?
> 
> I made the choice to not have too many statistics displayed in the UI, and I think it made more sense between 5 minutes and one hour to keep the second one.

hi @tcitworld :wave: 
On large systems, it is pretty easy to correlate the number of concurrent users to the resource consumption, which can be useful when one has to troubleshoot performance issues.
In my experience, the 5 minutes interval is pretty close to the exact number of concurrent users, whereas the last 1 hour indicates a way higher number.
So I would have kept the last 5 minutes :smiley: 

Do you know if there is another way to monitor the number of concurrent users or if it would be possible to add again the last 5 minutes indicator ?

tcitworld:
hi @maximelehericy
It's still available through the JSON API endpoint mentioned on the bottom of the page (under `activeUsers > last5minutes`).

I don't mind much about which one is selected for the UI, [I just picked one](https://github.com/nextcloud/serverinfo/pull/691#issuecomment-2394194081) between this one and the « one hour » value.
It's easy to change indicators shown in the UI, feel free to pick between the existing measurements.
* last 5 minutes
* last 1 hour
* last 24 hours
* last 7 days
* last 1 month
* last 3 months
* last 6 months
* last year

I'd keep a number of 4 shown in the UI to keep it not too crowded.

PR #723, opened 2024-12-12
chore(deps): drop PHP 8.0

kesselb:
Fix https://github.com/nextcloud/serverinfo/issues/721

Server requires at least PHP 8.1.


kesselb:
Merged via https://github.com/nextcloud/serverinfo/pull/725

PR #728, opened 2025-01-07
feat(storage): add size of appdata folder to statistics

tcitworld:
App data is used for a lot of things and the preview folder especially can grow very much. This helps identifying where storage growth comes from.

Also refactor to use `IAppConfig` in related files.


nickvergessen:
Please try the maintainers: https://github.com/nextcloud/serverinfo/blob/master/.github/CODEOWNERS#L1

tcitworld:
@JuliaKirschenheuter @kesselb why are you code owners only on the `/appinfo/info.xml` file and not on everything? I didn't have the reflex to check that.

come-nc (approved):
filecount would be useful as well, no?
On my personal instance the issue with previews was the number of files cluttering oc_filecache because of leftover previews.

tcitworld:
I can add this, yes.

kesselb (approved):

PR #733, opened 2025-01-17
fix(freebsd): fix cpu count for load graph

ConstrictM:
**Nextcloud version:** 30.0.5

FreeBSD is showing "CPU info not available". The reason is the CPU count function is not calling the command correctly.

Fix Tested:
image

### DIFF
--- a/lib/OperatingSystems/FreeBSD.php
+++ b/lib/OperatingSystems/FreeBSD.php
@@ -79,8 +79,9 @@ class FreeBSD implements IOperatingSystem {
 		$numCpu = -1;
 
 		try {
-			$numCpu = intval($this->executeCommand('sysctl -n hw.ncpu')); //TODO: this should be tested if it actually works on FreeBSD
+			$numCpu = intval($this->executeCommand('/sbin/sysctl -n hw.ncpu'));
 		} catch (RuntimeException) {
+			return $numCpu;
 		}
 
 		return $numCpu;


lib/OperatingSystems/FreeBSD.php
```diff
@@ -79,8 +79,9 @@ public function getCpuCount(): int {
 		$numCpu = -1;
 
 		try {
-			$numCpu = intval($this->executeCommand('sysctl -n hw.ncpu')); //TODO: this should be tested if it actually works on FreeBSD
+			$numCpu = intval($this->executeCommand('/sbin/sysctl -n hw.ncpu'));
 		} catch (RuntimeException) {
+			return $numCpu;
```
joshtrichards, line 84:
```suggestion
```

Should be unnecessary since we're pre-initializing `$numCpu` and returning it anyhow at the end of the function, but also doesn't hurt. I see we're doing this elsewhere already so I guess leave it.

joshtrichards (approved):
Thanks!

lib/OperatingSystems/FreeBSD.php
diamondpete: That was basically my thought process. Either remove all of the returns from the exception or at least make this function consistent with the other ones. Either way the value would be returned.

ConstrictM: That was my thought process. It should be consistent with the rest. Or all of the exception returns should be removed.

joshtrichards:
Can you edit your commit message? We're using conventional commits through more of our repositories these days. The requirement was recently activated here in this one too.

https://www.conventionalcommits.org/en/v1.0.0/#summary

joshtrichards (approved):

PR #738, opened 2025-01-23
Show php-fpm data if available

come-nc:
Adds this section below PHP, if php-fpm is used:
Copie d'écran_20250123_171216

### DIFF
--- a/lib/FpmStatistics.php
+++ b/lib/FpmStatistics.php
@@ -1,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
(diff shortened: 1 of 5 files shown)


lib/FpmStatistics.php
```diff
@@ -0,0 +1,50 @@
...
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\ServerInfo;
+
+use OCP\IDateTimeFormatter;
+
+/**
+ * @package OCA\Survey_Client\Categories
```
kesselb, line 15: Please remove it ;)

kesselb (changes requested):

kesselb (approved):
:elephant:

PR #743, opened 2025-01-28
Fix serverinfo representation in nextcloud OCC command tool: add info…

### LINKED ISSUE #190: Feature req: serverinfo representation in nextcloud OCC command tool
It would be a great option if I can get  serverinfo  output in the nextcloud OCC command tool.
It can be more useful for generating Munin graphs.

Yes, I know that, serverinfo has an external monitoring tool, which can be used for this, but my problem is, serverinfo  requires authentication, (Which is OK, when I'm using over the web.) but I should  update and manage login credentials in munin (or other) script. 

When I'm a root user on the server, I can run occ command without any nextcloud specific username and password, so it would be a great and helpful command in occ tool set.

And also it would be interesting to see in serverinfo:
How many users are logged in by web page.
How many users are connected with desktop or mobile app.
(In last 5min, 1hour, 24hour.)

acomdigital:
This PR adds a new OCC command, as described in [issue #190](https://github.com/nextcloud/serverinfo/issues/190), to retrieve serverinfo data.
Fixes #190 

### Changes Made:
- Added occ command `php occ serverinfo:info`;
- Encapsulated some business logic of ApiController into a dedicated service layer

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -26,6 +26,7 @@
 	</background-jobs>
 	<commands>
 		<command>OCA\ServerInfo\Commands\UpdateStorageStats</command>
+        <command>OCA\ServerInfo\Commands\GetInfo</command>
 	</commands>
     <settings>
         <admin>OCA\ServerInfo\Settings\AdminSettings</admin>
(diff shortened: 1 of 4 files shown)


JuliaKirschenheuter:
Hi @acomdigital!

Thank you for your pull request! Would you like to sfinish small thing which @kesselb suggested? Do you have an idea how to handle it?

PR #785, opened 2025-04-03
chore: add override attribute

kesselb:


nickvergessen:
Or instead of ignoring you add it? Worked for spreed, should work here

kesselb:
> Or instead of ignoring you add it? Worked for spreed, should work here

Thanks for the reminder, I didn't have the chance to finish this pr :see_no_evil:

Issue #809, opened 2025-05-15
Inaccurate statistics for active usage


akhil1508:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

### Steps to reproduce
1. Use nextcloud with a client that doesn't support cookies for e.g. a webdav mount: `mount -t davfs -o noexec,uid=$USER $NEXTCLOUD_DAV_URL_FOR_YOUR_USER webdav-folder`

### Expected behaviour
- This should be counted as an active login in last 30 days

### Actual behaviour
- It is not counted as an active login in last 30 days

### Investigation/suggestions
- This is likely because session tokens are not created [when a client doesn't support cookies](https://github.com/nextcloud/server/blob/f82ff164c3029c9b5a6c095563aff3b6b06cc438/lib/private/User/Session.php#L429)
- And the `serverinfo` app uses the [`authtokens` table](https://github.com/nextcloud/serverinfo/blob/0ba86031173c52b2db2ced5114da89f2d508db97/lib/SessionStatistics.php#L63) to get usage data
- Why not check the `lastLogin` timestamp instead which is stored in the `preferences` table and [updated on each new login](https://github.com/nextcloud/server/blob/f82ff164c3029c9b5a6c095563aff3b6b06cc438/lib/private/User/Session.php#L340)

### Server configuration

**Web server:** Nginx

**Database:** Maria

**PHP version:** 8.3

**Nextcloud version:** (see Nextcloud admin page) 30.0.4

<details>
<summary>List of activated apps</summary>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your Nextcloud installation folder
```
</details>

<details>
<summary>Nextcloud configuration</summary>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder
```
</details>

### Browser

**Browser name:** Firefox/Chrome/Safari/…

**Browser version:** 124/125/…

**Operating system:** Windows/Ubuntu/Mac/…

<details>
<summary>Browser log</summary>

```
Insert your browser log here, this could for example include:
a) The javascript console log
b) The network log
c) ...
```

</details>

PR #814, opened 2025-05-15
fix: get actual number of active users

hamza221:
Fix https://github.com/nextcloud/serverinfo/issues/809

The existing logic checks the `authtoken` table which comes with the tradeoff that user's row is deleted on logout.


kesselb:
Thank you :pray: 

I will review and test asap. 

@tcitworld what do you think?

hamza221:
I'll check the failing oracle test tomorrow

tcitworld (approved):
Funny enough, I've come to the same realization on the very same day this PR comes out.

Worst thing is probably people using WebDAV sessions which are deleted from `oc_authtoken` after one day.

PR #981, opened 2026-04-28
feat(services): add cron and background-job services

karlitschek:
Adds three read-only service classes used by the upcoming admin dashboard cards. None are wired into existing controllers yet, so this change is a pure addition with no behavior change.

* CronInfo      - cron mode (cron / webcron / ajax), last-run
                  timestamp, and a status threshold
* JobQueueInfo  - total / reserved / stuck job counts and the top
                  job classes by queue size
* SlowestJobs   - average and max execution time per job class

### DIFF
--- a/lib/CronInfo.php
+++ b/lib/CronInfo.php
@@ -1,0 +1,66 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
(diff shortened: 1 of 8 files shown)


ChristophWurst:
Fixed the original commit DCO and pushed the inclusion of the new classes into the initial state to avoid dead code.

ChristophWurst:
Fixed the generic catch-all to log the exception and added unit test coverage

ChristophWurst:
Squashed to clean up the history after the fixups

ChristophWurst (approved):

PR #982, opened 2026-04-28
feat(services): add log and login/session services

karlitschek:
Adds four read-only service classes used by the upcoming admin dashboard cards. None are wired into existing controllers yet, so this change is a pure addition with no behavior change.

* LogTailReader     - last N WARN/ERROR entries from the JSON log
                      (no shell, line-bounded)
* LoginStats        - bruteforce attempt counters and the top
                      offending IPs over recent windows
* ActivityRate      - oc_activity row counts over the last hour /
                      day / week
* ActiveConnections - session counts derived from oc_authtoken

### DIFF
--- a/lib/ActivityRate.php
+++ b/lib/ActivityRate.php
@@ -1,0 +1,87 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\ServerInfo;
+
+use OCP\App\IAppManager;
+use OCP\IDBConnection;
+
+class ActivityRate {
+	public function __construct(
+		private IAppManager $appManager,
+		private IDBConnection $db,
+	) {
+	}
+
+	/**
+	 * Counts user-facing activity events from the activity app over
+	 * recent time windows. Useful as a "is anything happening on the
+	 * server right now" signal.
+	 *
+	 * @return array{
+	 *     installed: bool,
+	 *     last1h: int,
+	 *     last24h: int,
+	 *     last7d: int,
+	 *     topActions: list<array{action: string, count: int}>
+	 * }
+	 */
+	public function getActivityRate(): array {
+		if (!$this->appManager->isInstalled('activity')) {
+			return ['installed' => false, 'last1h' => 0, 'last24h' => 0, 'last7d' => 0, 'topActions' => []];
+		}
+
+		try {
+			return [
+				'installed' => true,
+				'last1h' => $this->countSince(time() - 3600),
+				'last24h' => $this->countSince(time() - 86400),
+				'last7d' => $this->countSince(time() - 7 * 86400),
+				'topActions' => $this->topActions(),
+			];
+		} catch (\Throwable) {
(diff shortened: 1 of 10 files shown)


lib/ActivityRate.php
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	/**
+	 * @return list<array{action: string, count: int}>
+	 */
+	private function topActions(int $limit = 5): array {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('subjectparams', 'type')
+			->selectAlias($qb->func()->count('activity_id'), 'count')
+			->from('activity')
```
ChristophWurst, line 71: This could bite us: https://docs.nextcloud.com/server/latest/admin_manual/configuration_database/splitting.html#initial-splitting

ChristophWurst:
Activity, auth token and rate limiting are conceptually problematic. This creates an inter app dependency where this app accesses the persistence of another app directly. The technically sound way would be to go through a public API.

Looking into these right now …

ChristophWurst:
Amended the original commit to fix the DCO and added class usage like it will be used later to avoid dead code.

ChristophWurst (commented):
👍 functionality-wise

Needs decision conceptually

lib/LoginStats.php
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	/**
+	 * @return list<array{ip: string, count: int}>
+	 */
+	private function topIps(int $limit = 5): array {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('ip')
+			->selectAlias($qb->func()->count('id'), 'count')
+			->from('bruteforce_attempts')
```
kesselb, line 92: This only works for instances without a redis as distributed cache. 

https://github.com/nextcloud/server/blob/4eb8bc8d50e3f300aeb9b42a3c210d928b6a3709/lib/private/Server.php#L733-L750

lib/LogTailReader.php
```diff
@@ -0,0 +1,121 @@
...
+	 * with severity >= $minLevel (default: WARN = 2). Skips DEBUG/INFO.
+	 *
+	 * @return array{
+	 *     entries: list<array{time: string, level: int, app: string, message: string}>,
+	 *     available: bool,
+	 *     reason?: string
+	 * }
+	 */
+	public function recentErrors(int $limit = 8, int $minLevel = 2): array {
+		$logType = $this->config->getSystemValue('log_type', 'file');
```
kesselb, line 31: Should use ILogFactory / IWriter to fetch the entries.

Similar to the logreader app https://github.com/nextcloud/logreader/blob/2c83bb0a871a31378ac8967cce10ac613ca8e057/lib/Log/LogIteratorFactory.php#L29-L47

lib/Settings/AdminSettings.php
```diff
@@ -69,6 +77,10 @@ public function getForm(): TemplateResponse {
 			'cron' => $this->cronInfo->getCronInfo(),
 			'jobQueue' => $this->jobQueueInfo->getJobQueueInfo(),
 			'slowestJobs' => $this->slowestJobs->getSlowestJobs(),
+			'logTail' => $this->logTailReader->recentErrors(),
```
kesselb, line 80: @ChristophWurst 

Could we not bind them in getFrom but introduce a new controller (maybe ocs)?

 This has the advantage to not block the main operation with slow operations (like expensive database queries, opening log files, etc) and also give us the ability to let the dashboard refresh without a page load in the long run. 

Also okay as follow up, I would just mention it.

PR #983, opened 2026-04-28
feat(services): add storage, cache, and DB services

karlitschek:
Adds five read-only service classes used by the upcoming admin dashboard cards. None are wired into existing controllers yet, so this change is a pure addition with no behavior change.

* DbHealth             - largest tables on MySQL / PostgreSQL
* CachingInfo          - OPcache and APCu hit rates plus Redis config
* ExternalStoragesInfo - mounts configured via files_external
* DiskGrowth           - rolling history snapshots with a
                         days-until-full prediction
* TopUsersByQuota      - top users ranked by home::*/files size


lib/DbHealth.php
```diff
@@ -0,0 +1,99 @@
...
+	 * @return list<array{name: string, rows: int, sizeBytes: int}>
+	 */
+	private function mysqlTables(int $limit = 8): array {
+		$dbName = (string)$this->config->getSystemValue('dbname', '');
+		if ($dbName === '') {
+			return [];
+		}
+		$sql = 'SELECT table_name AS name, table_rows AS rows, '
+			. '(data_length + index_length) AS size_bytes '
+			. 'FROM information_schema.TABLES WHERE table_schema = ? '
```
ChristophWurst, line 59: This definitely needs error handler because by default Nextcloud drops admin privilege by creating a dedicated user during setup.

Checking options …

ChristophWurst: It should work when mariadb allows seeing the objects the user has access to. The only remaining issue is that the information_schema access is notorious for rebuilding database metadata.

Accepting the risk for now to get this merged.

lib/DbHealth.php
```diff
@@ -0,0 +1,99 @@
...
+			. 'ORDER BY size_bytes DESC LIMIT ' . (int)$limit;
+		$conn = $this->db;
+		$stmt = $conn->prepare($sql);
+		$stmt->bindValue(1, $dbName);
+		$result = $stmt->executeQuery();
+		$out = [];
+		while (($row = $result->fetch()) !== false) {
```
kesselb, line 66: Prefer using fetchAssociative/fetchNumeric/fetchOne or iterateAssociate/iterateNumeric instead.

lib/DbHealth.php
```diff
@@ -0,0 +1,99 @@
...
+		}
+		$result->closeCursor();
+		return $out;
+	}
+
+	/**
+	 * @return list<array{name: string, rows: int, sizeBytes: int}>
+	 */
+	private function pgTables(int $limit = 8): array {
+		$sql = 'SELECT relname AS name, n_live_tup AS rows, '
```
kesselb, line 81: Neither pg_class or pg_namespace have a column "n_live_tup".

I'm going to check ...

kesselb:
```sql
SELECT
    relname as name,
    n_live_tup as rows,
    pg_total_relation_size(relid) AS size_bytes
FROM pg_stat_user_tables
ORDER BY size_bytes DESC
LIMIT 8;
```

Image

kesselb: > pg_stat_user_tables

Should list all tables the current user has access too

lib/DiskGrowth.php
```diff
@@ -0,0 +1,128 @@
...
+	 *     hasEnoughData: bool
+	 * }
+	 */
+	public function getGrowthInfo(): array {
+		$history = $this->loadHistory();
+		$now = time();
+		$shouldSample = $history === [] || ($now - $history[count($history) - 1]['ts']) >= self::SAMPLE_INTERVAL;
+
+		if ($shouldSample) {
+			$dataDir = (string)$this->config->getSystemValue('datadirectory', '');
```
kesselb, line 53: I believe the sampling should go to https://github.com/nextcloud/serverinfo/blob/d0ea456f9d9d21df2847557d643c787020c1d19a/lib/StorageStatistics.php#L57 (which is called by background job every 3 hours)

Issue #551, opened 2024-01-10
Add new mechanism to get thermal notification


ostasevych:
Basically the serverinfo gets the information from /sys/class/thermal/thermal_zone*/temp.
At the same time some AMD motherboards and their chipsets do not store the information there, but in hwmon.
Eg, I have hp microserver, and it grabs and stores the temperature data :

```
k10temp:
temp1 /sys/devices/pci0000:00/0000:00:18.3/hwmon/hwmon3/temp1_input

w83795adg-i2c-1-2f:
temp1 /sys/devices/pci0000:00/0000:00:14.0/i2c-1/1-002f/temp1_input
temp2 /sys/devices/pci0000:00/0000:00:14.0/i2c-1/1-002f/temp2_input
temp5 /sys/devices/pci0000:00/0000:00:14.0/i2c-1/1-002f/temp5_input

jc42-i2c-0-18
temp1 /sys/devices/pci0000:00/0000:00:14.0/i2c-0/0-0018/hwmon/hwmon0/temp1_input

jc42-i2c-0-19
temp1 /sys/devices/pci0000:00/0000:00:14.0/i2c-0/0-0019/hwmon/hwmon0/temp1_input 
```

And

```
# ls -l /sys/class/hwmon/hwmon*/temp*_input
-r--r--r-- 1 root root 4096 січ 10 10:26 /sys/class/hwmon/hwmon0/temp1_input
-r--r--r-- 1 root root 4096 січ 10 10:26 /sys/class/hwmon/hwmon1/temp1_input
-r--r--r-- 1 root root 4096 січ 10 10:26 /sys/class/hwmon/hwmon3/temp1_input

# ls -l /sys/class/i2c*/i2c*/*/temp*_input
-r--r--r-- 1 root root 4096 січ 10 10:26 /sys/class/i2c-adapter/i2c-1/1-002f/temp1_input
-r--r--r-- 1 root root 4096 січ 10 10:26 /sys/class/i2c-adapter/i2c-1/1-002f/temp2_input
-r--r--r-- 1 root root 4096 січ 10 10:26 /sys/class/i2c-adapter/i2c-1/1-002f/temp5_input

```
As well as, lm-sensors produces good data.

Is that possible to grab the data in a more universal way, eg from `hwmon` class, but not from the `thermal_zone` class?

kesselb:
Hey, 

Using `/sys/class/hwmon/hwmon` looks okay to me. 

> The data are not so comfortable to interpret:

[...]

PR #1007, opened 2026-06-03
Vue

kesselb:


kesselb:
@coderabbitai review

ChristophWurst (commented):
A .nextcloudignore should be used to exclude source and package manager files from shipping, right?

ChristophWurst:
Was the second commit placed here intentionally?

kesselb:
> A .nextcloudignore should be used to exclude source and package manager files from shipping, right?

It's not necessary, because the build script covers all cases we have here. I'm adding still, because then it's explicit what should be in the release and what not. 

> Was the second commit placed here intentionally?

Yes, otherwise I won't see the sensors on my device ;) It also fixes https://github.com/nextcloud/serverinfo/issues/551.

ChristophWurst (approved):

benjaminfrueh (approved):

PR #1073, opened 2026-08-08
Drop custom grid & Add vue component library

kesselb:
## This PR

- Removes the custom grid in favor of CSS grid
- Introduces the vue component library
- Migrations the external monitoring section to NcFormGroup. 

### Example 1

Using CSS grid and removal of exceptions for components ensure a clean look across all components. 

The gap between cpu/memory, disk and network interfaces is the same. 

image

### Example 2

Using NcFormGroup adds a clean look, removes custom css, and adds the ability to copy the url or command. In addition a strong token is suggested. 

image


ChristophWurst (approved):
Not able to review this 🙈 

🐘

ChristophWurst (approved):
Actually … looks sane. Huge diff is bloated by lock file update

PR #1074, opened 2026-08-08
Resource usage overview & Tiles

kesselb:
## This PR

- Adds resource usage overview pane
- Introduces tiles for chart legends and loose stats
- Dividers for the overview tables
- Temp section moved down

### Here

Screenshot From 2026-08-08 20-45-09

### Main

Screenshot From 2026-08-08 20-44-27


ChristophWurst (approved):
🐘

PR #1084, opened 2026-08-09
Background jobs

kesselb:
## This PR

- Reworks the cron info, background jobs and slowest job (using IJobRuns to obtain failed and latest jobs, background job to collect the slowest jobs)
- Adds background jobs pane
- Tweaks the h3 style (so it's actually smaller than h2 and not bigger)

### Background jobs pane

image

image

image

### H3 tweak

| B | A |
| --- | --- |
| image | image |


ChristophWurst (approved):
🐘

PR #1122, opened 2026-09-09
fix: show when the page has stopped hearing from the server

karlitschek:
## Problem

Both composables handled their errors and then discarded the result.

`useStaticData` computed `loading` and `error`, but `SettingsAdmin.vue` destructured only `data` — so both flags were dead code. `useLiveData`'s `catch` block was empty by design. Every failure therefore looked identical from the outside, and none of them looked like a failure:

- a failed `/data` request left the pulsing skeletons on screen **forever**, with no error, no explanation, and no way to retry short of reloading the page
- a server that had been unreachable for an hour was **still polled every two seconds**, with the charts showing their last values and nothing saying those values had stopped moving

For a page whose entire job is reporting the health of a server, quietly showing stale numbers or a loading animation is the worst of the available failure modes: it looks healthy. An admin checking whether their server is alright would be reassured by a page that had lost contact with it.

## Change

- **A failed `/data` request says so**, and offers a **Try again** button that re-runs the request. The skeletons are no longer shown in that state, since nothing is loading.
- **Once two polls in a row fail** — one is usually a blip worth riding out — a warning names the time the values were last current: *"These values stopped updating at 17:15:42: the server is not responding."* The values stay on screen, which is still the right call; what changes is that the page stops implying they are live.
- **Polling backs off while requests are failing**, doubling from 2 s up to a minute and snapping back to normal on the first success, so a struggling server is not asked for the same numbers thirty times a minute.

## Verification

Driven against a live NC 35 instance with Playwright, failing the requests at the network layer:

| scenario | result |
| --- | --- |
| `/data` fails from the start | error card shown, skeletons gone, **"The server information could not be loaded. Try again"** |
| **Try again** clicked once the request is allowed through | data loads normally |
| page loads cleanly | **0** notices — no false positives |
| `/update` and `/periodic` start failing after a good load | **"These values stopped updating at 5:15:42 PM: the server is not responding."** |
| request attempts during 20 s of failure | **1**, where the 2 s interval would have made ~10 |

Screenshots of both states are in the branch discussion below. `eslint` clean; `js/` and `css/` rebuilt and committed, and a second build is a no-op.

## Note on overlap

This touches `useLiveData.ts` and `SettingsAdmin.vue`, so it will conflict textually with #1120 (which adds visibility gating to the same poller) and #1119 (which turns `SettingsAdmin.vue` into a shell). The changes are complementary rather than contradictory — I have them merged and running together on a test instance — so whichever lands first, I am happy to rebase the others onto it.


karlitschek:

kesselb (approved):

PR #1124, opened 2026-09-09
feat(database): warn about failing database checks on the overview page

karlitschek:
> **Stacked on #1123** (which is itself stacked on #1119), because it surfaces the checks that PR adds. GitHub retargets automatically as each one merges.

## Why

The database page reports what is wrong with the database configuration — but only to an admin who knows to open it. **Settings → Overview** is where they already look for things that need attention, so a failing check now says so there and links straight to the findings.

The result on a stock Ubuntu MariaDB:

> ⚠ **Database configuration**
> 12 database checks are failing. [Show the findings…](#)

## Not run on every page load

Running fifty-odd rules against a live server every time somebody opens Overview is not reasonable, so the outcome of a run is recorded in app config and the setup check reads that:

- **opening the database page records a fresh result**, so acting on a finding clears the warning without waiting for anything
- a server nobody has looked at for a day re-runs the checks once and records that instead
- anything that cannot be probed returns `info` rather than a scary warning — a database that is unreachable is a problem the other setup checks report far better than this one can

## Two things worth knowing

**`SetupResult::linkToDoc` is not rendered.** I passed the URL there first and it silently never appeared. Of the twenty-two setup checks on a stock install, exactly one renders a link — core's own code-integrity check — and it does it by passing a `highlight` rich object in `descriptionParameters` with `{link}` in the description. This now does the same. (DB Doctor's own setup check pastes a bare URL into the description text, which is presumably the same discovery arrived at differently.)

**The page had to become addressable.** A link to `#database` only works if the page reads it, so `SettingsAdmin.vue` now takes its view from the URL fragment (`#status`, `#database`, `#background-jobs`) and keeps it in sync with `history.replaceState` — `replaceState` rather than assigning to `location.hash`, so switching pages does not fill the back button with entries inside one settings section. An unrecognised fragment falls back to Status rather than rendering nothing.

## Verification

End to end against a live NC 35 instance:

- the warning renders in **Security & setup warnings** as *"Database configuration — 12 database checks are failing. Show the findings…"*
- the anchor is real: `<a href="/nextcloud/index.php/settings/admin/serverinfo#database">Show the findings…</a>` — 2 of 22 checks now render a link, where before it was 1
- following it lands with `hash="#database"`, the database view visible and the other two hidden
- switching tabs afterwards keeps the fragment in sync (`#background-jobs`)

`eslint`, `php-cs-fixer` and psalm clean; phpunit green; committed `js/`/`css/` reproduce from a fresh build.

## One judgement call to check

You asked for a **warning** whenever a check is failing, and that is what this does — any failing check, whatever its severity. On the test instance all 12 failures are `notice`-level tuning suggestions, and they produce a yellow warning. If that turns out to be noisier than you want on a typical install, the obvious refinement is `warning` when an alert or warning fails and `info` when only notices do. Easy to change — I did not want to quietly soften what you asked for.


karlitschek:

PR #1129, opened 2026-09-09
feat(database): add a database checks page

karlitschek:
> **Replaces #1123**, which went stale when #1119, #1120 and #1122 landed on master. Same work, rebuilt on current master with the conflicts resolved. Close #1123 in favour of this one.

The whole database feature in one branch, now that #1124, #1125 and #1128 have been merged into it: the page, the Overview setup check, the regression fixes and the rule pruning.

## What it adds

A **Database** page between Status and Background jobs, running the checks ported from Nextcloud's [DB Doctor](https://github.com/nextcloud/dbdoctor):

- **37 checks** grouped by category, run once when the page is first opened — no button, and results survive switching tabs
- three live boxes above them — **current connections, throughput, database latency** — polled only while the page is on screen and the tab is visible
- **read-only**: failing checks offer the config snippet to paste into `my.cnf` or `postgresql.conf`, but nothing writes to the database. A bundled monitoring app has no business holding write access to what it monitors
- an optional **separate connection** for the checks, so they can run as a user with rights Nextcloud itself does not need
- a **warning on Settings → Overview** when checks fail, linking to `#database`, cached so it does not re-run fifty rules on every page load

The phpMyAdmin attribution and GPL-2.0-or-later provenance travel with the ported rule set.

## What it deliberately leaves behind

The score, grade, mascot, history and trend charts — this is a page about what is wrong, not a dashboard to score. And 20 of the original 57 checks: the schema-integrity ones Nextcloud already reports on Overview, everything about the query cache (removed in MySQL 8.0) and MyISAM (Nextcloud runs on InnoDB), and findings an admin cannot act on such as short uptime or `slow_launch_time`.

Dropping the schema rules removed the only caller of the code that dispatches `AddMissingIndicesEvent` and friends to every installed app — by far the most expensive thing a check run did — and with it the dependency on `OC\DB\SchemaWrapper`, one fewer private server class for a bundled app to rely on.

## Conflict resolution, for review

Four files needed real merging against current master, and it is worth knowing which side won:

| File | Resolution |
| --- | --- |
| `StatusView.vue` | **master's** — #1122's `staticError` handling; the database work never touched it |
| `SettingsNavigation.vue`, `types.ts` | **this branch's** — a strict superset, adding the Database entry and types |
| `ConfigLexicon.php` | **union** — master's `cached_boot_time` entries plus `cached_db_check` |
| `SettingsAdmin.vue` | **union** — master's failure notices from #1122, plus the database page and the URL-fragment routing the Overview link needs |

Built bundles were regenerated rather than merged.

## Verification

Deployed to a live NC 35 instance with everything merged together:

- all three pages render; navigation shows Status / Database / Background jobs
- `/database/check` returns 200 with 37 rules, 5 failing, and the page renders 5 cards
- the Status page still shows the #1120 work correctly — 2 threads, real uptime, and no `NaN`, `-1024`, `1.0 B` or `(-1 threads)` anywhere
- the Overview warning tracks the new count

`eslint`, `php-cs-fixer` and psalm clean; phpunit green (37 tests in `LinuxTest` alone); the committed `js`/`css` reproduce from a fresh build.

## Still worth a maintainer's view

This duplicates DB Doctor's rule engine inside a bundled app, so the two copies will drift. I kept the ported files structurally identical and preserved upstream rule ids so a re-sync stays mechanical, but if you would rather serverinfo deferred to DB Doctor when it is installed, that is a better conversation to have now than after they diverge.


karlitschek:
Closing as redundant: #1123 was merged into master while I was preparing this, and I have verified the merge came out correct — master keeps #1122's failure notices *and* the database page in `SettingsAdmin.vue`, `StatusView.vue` keeps its `staticError` prop, and `ConfigLexicon` carries the `cached_boot_time` entries alongside `cached_db_check`.

The only remaining difference between this branch and master was the ordering of two entries in `ConfigLexicon`, which is cosmetic.

I also checked master directly: the committed `js`/`css` reproduce exactly from a fresh build (same chunk hash), eslint / php-cs-fixer / psalm are clean, and phpunit is green. Nothing here needs to land.

